PR #73, opened 2020-03-27
current date fix

AhdabNasir:


pages/all-data/AllData.html
```diff
@@ -177,12 +177,13 @@ <h3 class="govuk-heading-s panelTitle">Filters</h3>
     {% for tableFieldId in milestoneTableFields %}
     {% set flag = '' %}
     {% set field = milestone.fields.get(tableFieldId) %}
-      {% if (field.id == 'date') and (field.value < currentDate) %}
+      {% if (field.id == 'date') and (field.value < currentDate | date("YYYY-MM-DD")) %}
```
dtaylor7: actually can you change the model so date is coming back in the correct format?
`models/milestone.js` line: 43:
```
  date: {
    type: DATE,
    allowNull: false,
    displayName: 'Due Date',
    get() {
      if (!this.getDataValue('value')) return;
      return moment(this.getDataValue('value'), 'YYYY-MM-DD').format('DD/MM/YYYY');
    }
  }
```

AhdabNasir: The milestone format was fine, it was the currentDate that had the wrong format (date and time) so I've changed current date to match the milestone format. Hope that's ok :D

dtaylor7: yep ok fair enough 👍

PR #75, opened 2020-03-27
creating users with departments

gds-gharper:


scripts/createUser.js
```diff
@@ -1,49 +1,77 @@
...
+
+let email = readlineSync.question("Enter email: ");
+let departmentsString = readlineSync.question("Enter departments (seperated by ',', use 'all' for all departments): ");
 
-rl.question("Enter email: ", (email) => {
+let departments = [];
+if (departmentsString == "all") {
+  departments = getAllDepartments();
+} else {
+  departments = getDepartments(departmentsString);
```
dtaylor7: should this not be `await getDepartments(departmentsString);`

gds-gharper: You can't await in a non async function and since the script is running as base it's not.  Hence the promise check afterwards.

I could wrap the whole thing in an async function in order to accomplish this but this seemed a little nicer.

dtaylor7: it is an async func: https://github.com/alphagov/co-eu-transition-dashboard/pull/75/files#diff-d653eec9627c102ce309fd7d644a5cdbR10-R17

gds-gharper: The script itself isn't an async function, the easiest way for me to demonstrate is for you to add await in there and see what happens.

PR #77, opened 2020-03-27
block user if bad login

dtaylor7:
- block user if bad login
- fixed up unit tests for services/authentication
- add MySQL database migrations

### DIFF
--- a/services/authentication.js
+++ b/services/authentication.js
@@ -7,33 +7,42 @@ const User = require('models/user');
 const { Strategy: passportLocalStrategy } = require('passport-local');
 const { Strategy: passportJWTStrategy } = require("passport-jwt");
 const speakeasy = require('speakeasy');
+const maximumLoginAttempts = 3;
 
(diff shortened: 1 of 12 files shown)


services/authentication.js
```diff
@@ -102,7 +111,17 @@ const login = (req, res) => {
...
+      if (err.loginAttempts) {
+        if (err.loginAttempts === maximumLoginAttempts) {
+          req.flash(`You account is now locked, please contact us.`);
+        } else {
+          req.flash(`Incorrect username and/or password entered, ${maximumLoginAttempts - err.loginAttempts} login attempts remaining`);
+        }
+      } else {
+        req.flash(`Incorrect username and/or password entered`);
+      }
+
```
gds-gharper: It's a trade off between usability and security but in theory you should simply tell them their login failed rather than why it failed.

dtaylor7: done

PR #78, opened 2020-03-30
logout and email change

AhdabNasir:


test/unit/pages/Logout.test.js
```diff
@@ -0,0 +1,30 @@
...
+  beforeEach(() => {
+    const Logout = require('pages/logout/Logout');
+
+    res = { redirect: sinon.stub(), clearCookie: sinon.stub() };
+
+    const req = { cookies: [] };
+
+    page = new Logout('some path', req, res);
+
+    sinon.stub(jwt, 'restoreData');
```
dtaylor7: dont think we need this stub?

common/views/template.njk
```diff
@@ -54,6 +61,20 @@
       <a href="/" class="govuk-header__link govuk-header__link--service-name">
         Transition Delivery Dashboard
       </a>
+
+  {% if logout == "true" %}
```
dtaylor7: this could be `if req.user`?

AhdabNasir: I agree, that'd be a much better way of doing it, or checking if the session is active rather than by page.

AhdabNasir: I've changed the code slightly similar to what you were saying here

PR #79, opened 2020-03-31
DDB-116 - As a user, when i login for the first time i am required to change my password

dtaylor7:
[DDB-116 - As a user, when i login for the first time i am required to change my password](https://technologyprogramme.atlassian.net/browse/DDB-116)
 - If a user has not changed their password before allowing them two after completing 2FA


models/user.js
```diff
@@ -107,6 +107,10 @@ User.init({
   loginAttempts: {
     type: INTEGER,
     field: "login_attempts"
+  },
+  passwordReset: {
+    type: BOOLEAN,
+    field: "password_reset"
   }
```
gds-gharper: Normally you would name boolean fields with a classifier like is_changed, is_active, perhaps this one could be must_change_password ?

pages/authentication/password-reset/PasswordReset.js
```diff
@@ -0,0 +1,96 @@
...
+    const passwordHasNumbers = /\d/.test(password);
+    const passwordLowerCase = /[a-z]/.test(password);
+    const passwordUpperCase = /[A-Z]/.test(password);
+    const passwordSpecialCharacter = /[!@#$%^&*()_+\-=[\]{};':"\\|,.<>/?]+/.test(password);
+    const passwordsMatch = password === confirmPassword;
+
+    let error;
+
+    if (passwordLength < 8) {
+      error = 'The password must have a minimum of 8 charaters';
```
gds-gharper: This should probably be in config.  Also charaters => characters

dtaylor7:
> The other way of doing this rather than having a boolean field which forces a password reset would be to have a seperate field/table which stores a resetPassword. When that is set or they log in with that password then they are forced to change their password.
> 
> That way you can handle forgotten passwords too. So when we create a user we'd create them with the resetPassword and send that out via email. When they then try to log in we see if that field is set, check the password against that, if it matches, we force them to change their password and clear that field when it's done. If they forget their password we simply set that field with a new hashed password and send that via email, allowing them to log in (without resetting their original password until they actually log in with the reset password).

There are many ways to skin a cat. We can also achieve the above using the existing password field and this new boolean field.

gds-gharper:
The problem with doing it that way is when someone hits reset password they immediately lose their current password, allowing for DoS.  Having a separate field for password reset vs password means you don't clear their existing password when they request a reset.

dtaylor7:
> The problem with doing it that way is when someone hits reset password they immediately lose their current password, allowing for DoS. Having a separate field for password reset vs password means you don't clear their existing password when they request a reset.

We can decide on an approach when a password reset story is played

migrations/01_user_update_password.js
```diff
@@ -0,0 +1,24 @@
...
+
+module.exports = {
+  up: async (query) => {
+    // logic for transforming into the new state
+    await query.addColumn(
+      'user',
+      'must_change_password',
+      {
+        type: Sequelize.DataTypes.BOOLEAN,
+        defaultValue: 1
```
gds-gharper: Whilst we definitely want this to be defaulted to 1 (force reset), do we want to force all users to change their passwords who have already been set up (Probably yes, but worth preparing them for this).

dtaylor7: yes, sophie is already aware

PR #86, opened 2020-04-03
Missed milestones

dtaylor7:
Added missed milestones page

Developed by @AhdabNasir and me


pages/authentication/two-factor-authentication/TwoFactorAuthentication.js
```diff
@@ -27,7 +27,7 @@ class TwoFactorAuthentication extends Page {
     } else if (this.req.user.role === 'admin') {
       return this.res.redirect(config.paths.admin.import);
     } else {
-      return this.res.redirect(config.paths.allData);
+      return this.res.redirect(config.paths.missedMilestones);
```
gds-gharper, line 30: Given we keep having to change this in multiple places might be worthwhile creating a startPage function (since it's likely eventually we'll want different users hitting different start pages).

dtaylor7: sounds like a plan, can you add task to jira?

PR #91, opened 2020-04-03
Add upcoming milestone page

dtaylor7:


common/javascript/upcoming-milestones.js
```diff
@@ -0,0 +1,89 @@
...
+  const ctx = document.getElementById(elementId);
+  ctx.height = xAxisLablesHeight + (data.data.length * rowHeight * 2);
+
+  const addEllipsis = (string, maxLength = 30) => {
+    let text = string.substr(0, maxLength);
+    if(string.substr(maxLength, 1).length) {
+      text += '...';
+    }
+    return text;
+  }
```
gds-gharper, line 17: Could this be accomplished with css https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

common/javascript/upcoming-milestones.js
```diff
@@ -0,0 +1,89 @@
...
+
+  new Chart(ctx, {
+    type: 'bubble',
+    data: {
+      datasets: [{
+        radius: 10,
+        hoverRadius: 15,
+        data: data.data,
+        backgroundColor: '#5694CA',
+        hoverBackgroundColor: '#5694CA'
```
gds-gharper, line 27: Quite a lot of magic numbers in here which seem to be the same largely, would be nice to at least stick them in consts and give them names.

common/javascript/upcoming-milestones.js
```diff
@@ -0,0 +1,89 @@
...
+      scales: {
+        xAxes: [{
+          ticks: {
+            precision: 0,
+            min: data.min,
+            max: data.max,
+            callback: function(value) {
+              return moment(value, 'x').format('DD MMMM');
+            },
+            stepSize: 100000000 * 4,
```
gds-gharper, line 73: What does this number represent?

models/user.js
```diff
@@ -53,6 +54,16 @@ class User extends Model {
         ]
       }]
     });
+
+    // sort milestones inside projects
+    departmentsWithProjects.forEach(departments => {
+      departments.projects.forEach(projects => {
+        projects.milestones = projects.milestones
+          .sort((a, b) => (moment(a.date, 'DD/MM/YYYY').isAfter(moment(b.date, 'DD/MM/YYYY'))) ? 1 : -1);
```
gds-gharper, line 62: You could end up somewhat compromising the sort performance here, it's minor but you should return 0 when they're the same.  Technically it's possible to send the sort into an infinite loop if you get unlucky

common/javascript/upcoming-milestones.js
dtaylor7: this is controlled inside the chart.js library,  I can override the HTML to create the tooltip but this adds a lot of extra complexity.

common/javascript/upcoming-milestones.js
dtaylor7: @AhdabNasir could you take a look at this one?

common/javascript/upcoming-milestones.js
dtaylor7: have removed it

models/user.js
dtaylor7: done

common/javascript/upcoming-milestones.js
AhdabNasir: This has now been done :)

PR #103, opened 2020-04-08
Hide milestones features behind feature toggle

dtaylor7:
- add new start page ( used to redirect from the root )
 - hide missed milestones behind feature toggle
 - hide upcoming milestone chart behind a feature toggle

### DIFF
--- a/app.js
+++ b/app.js
@@ -6,6 +6,7 @@ const logger = require('middleware/logger');
 const helmet = require('middleware/helmet');
 const cookieParser = require('cookie-parser');
 const config = require('config');
+const startPage = require('helpers/startPage');
 
 const app = module.exports = express();
 
@@ -26,6 +27,6 @@ if (config.env === 'development') {
 logger.attachRouteLogger(app);
 
 pages.attach(app);
-app.get('/', (req, res) => res.redirect(config.paths.start));
+app.get('/', (req, res) => res.redirect(startPage()));
 
 logger.attachErrorLogger(app);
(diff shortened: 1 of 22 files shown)


app.js
```diff
@@ -26,6 +26,6 @@ if (config.env === 'development') {
 logger.attachRouteLogger(app);
 
 pages.attach(app);
-app.get('/', (req, res) => res.redirect(config.paths.missedMilestones));
+app.get('/', (req, res) => res.redirect(config.paths.start));
```
gds-gharper, line 29: Can we wrap this config.paths.start in a function instead of delving direct into the config hash.

That way when we want to move to role based features we can simply change the function rather than finding everywhere which calls config.  (For now it can just return config.paths.start)

dtaylor7: can you provide a code example? Not sure what you mean here?

gds-gharper: Something along the lines of

```
const startPage = require('startPage');
app.get('/', (req, res) => res.redirect(startPage()));
```

where the startPage function for now simply returns config.paths.start.  That means when we change to user based differing start pages we can handle it in that function, in one place, rather than having to change everywhere which delves into config.

You may wish to make the name a little more generic so the module can be used for other things than start page.

dtaylor7: that is effectively what I have done, moved all redirection logic to the start page

dtaylor7: where shall I put this new function?

gds-gharper: Create a new module for it.  You could place it inside the user model potentially since in the future this is likely to be based off user though that may be tying the user model to the end code and config.

PR #106, opened 2020-04-08
Require current password on change

gds-gharper:


pages/authentication/password-reset/PasswordReset.js
```diff
@@ -56,12 +56,26 @@ class PasswordReset extends Page {
       error = 'The passwords must match';
     }
 
+
```
dtaylor7, line 59: minor - remove extra space here

pages/authentication/password-reset/PasswordReset.js
```diff
@@ -56,12 +56,26 @@ class PasswordReset extends Page {
...
     return error;
   }
 
   async passwordReset() {
     const password = this.req.body.password;
     const confirmPassword = this.req.body['confirm-password'];
+    const currentPassword = this.req.body['current-password'];
+
+    let hasError = await authentication.authenticateLogin(this.req.user.email,currentPassword, (error,user) => {
+      if (!user || error) {
```
dtaylor7, line 69: 1. rename `hasError` to `errorWithCurrentPassword`

2. you don't need to check for user here as an error is returned if the user does not exist. 
so this can be rewritten as:
```
const errorWithCurrentPassword = await authentication.authenticateLogin(this.req.user.email, currentPassword, error => error);
if (errorWithCurrentPassword) {
  this.req.flash(`Incorrect password entered`);
  return this.res.redirect(config.paths.authentication.passwordReset);
}
```

also as an alternative, you could rewrite `authenticateLogin` function to check if `done` callback is passed and if not just return `[error, user]`;

pages/authentication/password-reset/PasswordReset.js
```diff
@@ -56,12 +56,26 @@ class PasswordReset extends Page {
...
+
     return error;
   }
 
   async passwordReset() {
     const password = this.req.body.password;
     const confirmPassword = this.req.body['confirm-password'];
+    const currentPassword = this.req.body['current-password'];
+
+    let hasError = await authentication.authenticateLogin(this.req.user.email,currentPassword, (error,user) => {
```
dtaylor7, line 68: this should be `const` as it isn't reassigned

pages/authentication/password-reset/PasswordReset.js
```diff
@@ -62,6 +62,12 @@ class PasswordReset extends Page {
   async passwordReset() {
     const password = this.req.body.password;
     const confirmPassword = this.req.body['confirm-password'];
+    const currentPassword = this.req.body['current-password'];
+
+    if (await authentication.authenticateLogin(this.req.user.email,currentPassword, error => error)) {
```
dtaylor7, line 67: It would be easier to understand and self-explanatory if you pulled this out into a variable above that described what it is I.e.
`const currentPasswordIsIncorrect = await authentication.authenticateLogin(this.req.user.email, currentPassword, error => error)`

dtaylor7:
All good, apart from small optional comment

pages/authentication/password-reset/PasswordReset.js
gds-gharper: I think it's fairly self explanatory tbh at this.

PR #108, opened 2020-04-14
Bulk upload

dtaylor7:
- Create parse helpers to parse data from excel bulk upload
- Create validation helpers to validate data from excel bulk upload
- Create a temporary bulk upload store to store data from bulk upload
- Add new page to either show errors for upload or enable user to finalise upload


helpers/utils.js
```diff
@@ -33,6 +39,16 @@ const removeNulls = (obj) => {
...
   return obj;
-}
+};
+
+const pluck = (collection, key) => {
+  return collection.reduce((keys, item) => {
+    keys.push(item[key]);
+    return keys;
+  }, []);
+};
```
gds-gharper, line 49: Isn't this just a single map `return collection.map(value => value[key])`

helpers/validation.js
```diff
@@ -0,0 +1,123 @@
...
+
+const validateDate = value => {
+  const isValidDate = moment(value).isValid();
+  if(!isValidDate) {
+    throw Error(`Is not a valid date`);
+  }
+};
+
+const validateInGroup = (value, options) => {
+  const valueInGroup = options.includes(value);
```
gds-gharper, line 31: to be consistent with the rest of the code shouldn't this be isValueInGroup?

migrations/03_bulk_upload.js
```diff
@@ -0,0 +1,63 @@
...
+    await query.createTable('temporary_bulk_import_store', {
+      id: {
+        type: Sequelize.STRING(40),
+        allowNull: false,
+        primaryKey: true
+      },
+      user_id: {
+        type: Sequelize.INTEGER,
+        allowNull: false
+      },
```
gds-gharper, line 14: This needs to be a foreign key

migrations/03_bulk_upload.js
```diff
@@ -0,0 +1,63 @@
...
+      created_at: {
+        type: Sequelize.DATE,
+        defaultValue: Sequelize.NOW
+      },
+      updated_at: {
+        type: Sequelize.DATE,
+        defaultValue: Sequelize.NOW
+      }
```
gds-gharper, line 26: If you look at the definitions of the other created at and updated at columns you'll see they have very specific declarations which means they automatically update the appropriate values at the appropriate times ```
CREATE TABLE `milestone` (
  `uid` varchar(32) NOT NULL,
  `project_uid` varchar(32) NOT NULL,
  `description` text,
  `date` date DEFAULT NULL,
  `created_at` timestamp NULL DEFAULT CURRENT_TIMESTAMP,
  `updated_at` timestamp NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  PRIMARY KEY (`uid`),
  KEY `fk_milestone_project_idx` (`project_uid`),
  CONSTRAINT `fk_milestone_project` FOREIGN KEY (`project_uid`) REFERENCES `project` (`uid`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4```

migrations/03_bulk_upload.js
```diff
@@ -0,0 +1,63 @@
...
+    });
+
+    // logic for transforming into the new state
+    await query.addColumn(
+      'milestone_field',
+      'mi_template_column_name',
+      {
+        type: Sequelize.DataTypes.STRING(255)
+      }
```
gds-gharper, line 35: I guess these should be unique

models/milestone.js
```diff
@@ -13,6 +16,76 @@ class Milestone extends Model {
...
+      miTemplateColumnName: 'Target date for delivery'
+    }]);
+
+    return milestoneFields;
+  }
+
+  static async import(milestone, milestoneFields, options) {
+    const attributes = Object.keys(milestone);
+
+    const milestoneAttribites = attributes.filter(attribute => this.rawAttributes[attribute]);
```
gds-gharper, line 51: s/attribites/attributes/

models/milestone.js
```diff
@@ -13,6 +16,76 @@ class Milestone extends Model {
...
+    const milestoneToUpsert = milestoneAttribites.reduce((milestoneToUpsert, attribute) => {
+      milestoneToUpsert[attribute] = milestone[attribute];
+      return milestoneToUpsert;
+    }, {});
```
gds-gharper, line 55: I see this pattern quite a bit in your code, you could use the pick command from lodash https://lodash.com/docs/#pick

migrations/03_bulk_upload.js
```diff
@@ -0,0 +1,63 @@
...
+    );
+
+    await query.addColumn(
+      'project_field',
+      'mi_template_column_name',
+      {
+        type: Sequelize.DataTypes.STRING(255)
```
gds-gharper, line 42: I'd probably call this import_column_name, same with all of them below

models/milestoneFieldEntry.js
```diff
@@ -2,8 +2,28 @@ const { Model, STRING, INTEGER, TEXT } = require('sequelize');
...
+    if (!milestoneFieldEntry.value) {
+      return
+    }
+
+    const exsistingField = await MilestoneFieldEntry.findOne({
+      where: {
+        milestoneUid: milestoneFieldEntry.milestoneUid,
+        fieldId: milestoneFieldEntry.fieldId
+      }
+    });
```
gds-gharper, line 18: There's potentially a race condition here if someone double clicks though it's probably minor, you can avoid it by doing an actual upsert (mysql syntax is INSERT ... ON DUPLICATE KEY ...)

models/milestoneFieldEntryAudit.js
```diff
@@ -0,0 +1,31 @@
...
+    primaryKey: true
+  },
+  value: {
+    type: TEXT
+  },
+  archivedAt: {
+    type: DATE,
+    field: 'archived_at',
+    defaultValue: NOW
+  }
```
gds-gharper, line 25: You're missing created_at and updated_at (which would be copied directly from the origin)

models/projectFieldEntry.js
```diff
@@ -2,12 +2,32 @@ const { Model, STRING, INTEGER, TEXT, Op } = require('sequelize');
...
+
+    const exsistingField = await ProjectFieldEntry.findOne({
+      where: {
+        projectUid: projectFieldEntry.projectUid,
+        fieldId: projectFieldEntry.fieldId
+      }
+    });
+
+    if (exsistingField) {
+      await ProjectFieldEntryAudit.create(exsistingField.toJSON(), options);
```
gds-gharper, line 25: s/exsisting/existing/

models/projectFieldEntryAudit.js
```diff
@@ -0,0 +1,31 @@
...
+    primaryKey: true
+  },
+  value: {
+    type: TEXT
+  },
+  archivedAt: {
+    type: DATE,
+    field: 'archived_at',
+    defaultValue: NOW
+  }
```
gds-gharper, line 25: Missing created_at and updated_at

models/temporaryBulkImportStore.js
```diff
@@ -0,0 +1,21 @@
+const { Model, INTEGER, JSON } = require('sequelize');
+const sequelize = require('services/sequelize');
+
+class TemporaryBuilImportStore extends Model {}
```
gds-gharper, line 4: s/Buil/Bulk/

models/temporaryBulkImportStore.js
```diff
@@ -0,0 +1,21 @@
...
+
+TemporaryBuilImportStore.init({
```
gds-gharper, line 6: I'd be tempted to lose the temporary and store (everything in a database is a store) from this name and just call it BulkImport

models/temporaryBulkImportStore.js
```diff
@@ -0,0 +1,21 @@
...
+  id: {
+    type: INTEGER,
+    field: 'id',
+    primaryKey: true
+  },
+  userId: {
+    type: INTEGER,
+    field: 'user_id'
+  },
```
gds-gharper, line 15: This needs a foreign key

helpers/utils.js
dtaylor7: yep, updated this

models/projectFieldEntry.js
dtaylor7: what does this mean?

dtaylor7: oh i see, spelling

migrations/03_bulk_upload.js
dtaylor7: the reality is here we will only have a created at so I have removed the updated at date

models/milestoneFieldEntry.js
dtaylor7: sequelize, does not currently support this in one query.

models/milestoneFieldEntryAudit.js
dtaylor7: this gets automatically added by sequalise `}, { sequelize, modelName: 'projectFieldEntryAudit', tableName: 'project_field_entry_audit', createdAt: 'created_at', updatedAt: 'updated_at' });`

models/projectFieldEntryAudit.js
dtaylor7: this gets automatically added by sequalise `}, { sequelize, modelName: 'projectFieldEntryAudit', tableName: 'project_field_entry_audit', createdAt: 'created_at', updatedAt: 'updated_at' });`

models/milestoneFieldEntry.js
gds-gharper: It has upsert and you'd use a compound primary/unique key of milestoneUid and fieldId

dtaylor7: the reason I am do the find first is so that if it exists I can create it in the MilestoneFieldEntryAudit table?

helpers/models.js
```diff
@@ -56,43 +45,7 @@ const transformForView = model => {
 };
 
 const truthy = [true, 'true', 'yes', 1, '1', 'y'];
```
gds-gharper, line 47: We've got this exact thing in helpers/validation as well, probably worth merging them

helpers/parse.js
```diff
@@ -0,0 +1,72 @@
...
+
+const parseDateExcel = (excelTimestamp) => {
+  const secondsInDay = 24 * 60 * 60;
+  const excelEpoch = new Date(1899, 11, 31);
+  const excelEpochAsUnixTimestamp = excelEpoch.getTime();
+  const missingLeapYearDay = secondsInDay * 1000;
+  const delta = excelEpochAsUnixTimestamp - missingLeapYearDay;
+  const excelTimestampAsUnixTimestamp = excelTimestamp * secondsInDay * 1000;
+  const parsed = excelTimestampAsUnixTimestamp + delta;
+  return isNaN(parsed) ? null : parsed;
```
gds-gharper, line 21: I think we'd be better off using a date library like moment to do these calculations rather than rolling our own.  Why are we subtracting 1 days worth of milliseconds from the timestamp?  That way you can also return a moment time from here as well.

https://github.com/SheetJS/sheetjs (which node-xslx uses) is supposed to handle dates itself.  If you pass in the option `cellDates: true` when you create the parser it should generate iso8601 format strings which you can then throw at moment.

migrations/03_bulk_upload.js
gds-gharper: Why are we not changing the names from "mi_template_", I'm very much not keen on mi_template_ which is why I suggested changing it to import_column_name (which is more descriptive of what it is).

gds-gharper: Sorry, was viewing an old diff on this comment

models/milestoneFieldEntry.js
gds-gharper: I guess it's fine for now, we can look an optimising it later on, I don't imagine we'll be hitting a race condition very often, if ever for now.

helpers/parse.js
dtaylor7: moment doesnt know how to handle excel date timestamps, it seems to be something unique to excel

dtaylor7: i will see how https://github.com/SheetJS/sheetjs is handling this

dtaylor7: ok all done

PR #113, opened 2020-04-16
Validate excel, parse excel document in correct format

dtaylor7:


test/unit/services/xlsx.test.js
```diff
@@ -2,23 +2,85 @@ const { expect, sinon } = require('test/unit/util/chai');
 const xlsx = require('xlsx');
```
gds-gharper, line 2: This is a good start, but I'd like some tests on an actual xls file read from the filesystem as well if possible.  Ideally a few, some from windows and mac.

dtaylor7: Would I be able to do that retrospectively so ahdab can start integrating with this?

gds-gharper: If you create a ticket for it then I'll approve.

dtaylor7: https://technologyprogramme.atlassian.net/browse/DDB-244

PR #121, opened 2020-04-20
only show none complete milestones in missed milestones

dtaylor7:
- speed up database queries
- only show none complete milestones in missed milestones

### DIFF
--- a/models/milestoneFieldEntry.js
+++ b/models/milestoneFieldEntry.js
@@ -1,4 +1,4 @@
-const { Model, STRING, INTEGER, TEXT } = require('sequelize');
+const { Model, STRING, INTEGER, TEXT, Op } = require('sequelize');
 const sequelize = require('services/sequelize');
 const modelUtils = require('helpers/models');
 const MilestoneField = require('./milestoneField');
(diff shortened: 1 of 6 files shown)


models/milestoneFieldEntry.js
```diff
@@ -33,6 +33,21 @@ class MilestoneFieldEntry extends Model {
...
     });
   }
+
+  static createSearch(milestoneField, options) {
+    const filters = options.map(option => {
+      return {
+        [Op.and]: {
+          milestone_field_id: milestoneField.id,
+          value: {
+            [Op.like]: `%${option}%`
```
gds-gharper, line 43: double ended like filters are incredibly inefficient (and can never hit indexes either).  You also have the problem where substrings in values will match as well.

I'd prefer it if we just did exact matches, any reason not to (other than allowing people to search for substrings)?

dtaylor7: this was here before to query blobs, but I guess now you have changed to TEXT we can query normally?

dtaylor7: have updated

gds-gharper: sure, if you can do that quickly then I'm much happier

PR #124, opened 2020-04-23
Admin list project fields

AhdabNasir:


pages/admin/project-field-list/ProjectFieldList.html
```diff
@@ -0,0 +1,89 @@
...
+  <hr class="govuk-section-break govuk-section-break--m">
+
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+
+  <table class="govuk-table project-fields">
+    <caption class="govuk-table__caption">Department and Project Information</caption>
```
dtaylor7, line 26: I have already mentioned this, but just put it on the review.

Can we use the dynamic data from the database? I have put a function in the step so groups are available to the template

PR #129, opened 2020-04-29
change latest two migrations so they can rollback if errors occur

dtaylor7:


config/development.json
```diff
@@ -8,7 +8,8 @@
   "services": {
     "mysql": {
       "uri": "mysql://root:password@127.0.0.1:3306/dashboard",
-      "sslCertificate": null
+      "sslCertificate": null,
+      "charset": "utf8"
```
gds-gharper, line 12: This should be utf8mb4 or you're going to have differences between dev and other environments.  The entire table type is `ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci`

dtaylor7: updated

PR #132, opened 2020-04-30
Adding safe filter

AhdabNasir:


common/views/feedback.njk
```diff
@@ -6,5 +6,5 @@
       tag: {
         text: "beta"
       },
-      html: "This is a new service – your <a class='govuk-link' href='" + feedbackLink + "'>feedback</a> will help us to improve it."
+      html: "This is a new service – your <a class='govuk-link' href='" + feedbackLink + "'>feedback</a> will help us to improve it." | safe
```
gds-gharper, line 9: Have you tested this, I strongly suspect it will encode the <a href> as well?

AhdabNasir: I tested it and it works fine

gds-gharper: Sorry, I misread `safe` as `escape`, does it actually encode feedbacklink?  You're telling it that string is safe, not telling it to escape the string, you should be using `escape`.  If you changed this to escape you would encode the whole string.

For instance if you set feedbackLink as so `set feedbackLink = "'><script>alert(\"Hello World\");</script>";`. Do you get an alert?

Please add a test for this such that you can see for sure that a dangerous string is properly escaped this way we can confirm that the story is correct and that we don't regress.

common/macros/critical-milestones-macro.njk
```diff
@@ -3,9 +3,9 @@
 
 {% macro criticalMilestones(projects, projectFields, milestoneFields, extraField, config) %}
 
-  {% set uid = '' %}
-  {% set cellBlock = "cell-color-" %}
-  {% set flag = '' %}
+  {% set uid = '' | escape %}
+  {% set cellBlock = "cell-color-" | escape %}
+  {% set flag = '' | escape %}
```
gds-gharper, line 8: You don't need to escape these because you're simply escaping an empty string (or in the second, the case of "cell-color-" neither of which escape anything.  Also you don't want to escape html attributes in the same way as html itself (in this case it probably won't escape speech marks).  Accordint go the documentation you should use jsonAttribute to escape attributes.  http://ohdotfive.apostrophenow.org/api-reference/nunjucks-filters.html

common/macros/critical-milestones-macro.njk
```diff
@@ -61,7 +61,7 @@
               {% for milestoneField in milestoneFields %}
                 {% set field = milestone.fields.get(milestoneField.id) %}
                 {% if (field.id == 'date') and (field.value | beforeToDay) %}
-                  {% set flag = 'red-date' %}
+                  {% set flag = 'red-date' | escape %}
```
gds-gharper, line 64: Same issue with escaping a static string here.

common/views/feedback.njk
```diff
@@ -1,10 +1,10 @@
     
     {% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
-    {% set feedbackLink = "https://docs.google.com/forms/d/19Ymg-262uTXzVW8C8bAKxX9cn3TMh-ItZH69C3mmIe0/edit?ts=5da6e72e" %}
+    {% set feedbackLink = "https://docs.google.com/forms/d/19Ymg-262uTXzVW8C8bAKxX9cn3TMh-ItZH69C3mmIe0/edit?ts=5da6e72e" | escape %}
```
gds-gharper, line 3: This is probably jsonAttribute escaping

pages/all-data/AllData.html
```diff
@@ -158,16 +158,16 @@ <h3 class="govuk-heading-s panelTitle">Filters</h3>
...
-    {% set uid = field.value %}
+    {% set uid = field.value | escape %}
     {% elif (field.id == 'departmentName') and (departmentView == 'true') %}
     {% else %}
 
     {% set projectValue = (projectValue.push({html:
 
     "<a href=" + config.paths.projectDetails | replace(":uid", uid | replace(" ", "%20"))  + " class='govuk-link project-link'>" +
-    "<div class='scrollable'>" + field.value + "</div></a>" | safe
+    "<div class='scrollable'>" + field.value + "</div></a>"
```
gds-gharper: This needs field.value escaping, would something like below work

```"<div class='scrollable'>" + ( field.value  |escape) + "</div></a>"```

 I don't know if you can apply filters within brackets.  If not you'll need to set a variable to the escaped vlaue of field.value and then set that.

File: pages/all-data/AllData.html [hunk 6]
```diff
@@ -182,14 +182,14 @@ <h3 class="govuk-heading-s panelTitle">Filters</h3>
     {% set milestoneValue = [] %}
 
     {% for tableFieldId in milestoneTableFields %}
-    {% set flag = '' %}
+    {% set flag = '' | escape %}
```
gds-gharper, line 185: Escaping an empty string accomplishes nothing

pages/all-data/AllData.html
```diff
@@ -158,16 +158,16 @@ <h3 class="govuk-heading-s panelTitle">Filters</h3>
...
     {% else %}
 
     {% set projectValue = (projectValue.push({html:
 
     "<a href=" + config.paths.projectDetails | replace(":uid", uid | replace(" ", "%20"))  + " class='govuk-link project-link'>" +
-    "<div class='scrollable'>" + field.value + "</div></a>" | safe
+    "<div class='scrollable'>" + field.value + "</div></a>"
       if field.id == 'title'
-        else "<span class=" + cellBlock + field.value + ">" + field.value | default('N/A') + "</span>" | safe
+        else "<span class=" + cellBlock + field.value + ">" + field.value | default('N/A') + "</span>"
```
gds-gharper: In this case you want to jsonEscape the value of `(cellBlock + field.value)` and html escape the value of `field.value | default('N/A')`. It's likely you'll have to put that into a variable

pages/all-data/AllData.html
```diff
@@ -182,14 +182,14 @@ <h3 class="govuk-heading-s panelTitle">Filters</h3>
     {% set milestoneValue = [] %}
 
     {% for tableFieldId in milestoneTableFields %}
-    {% set flag = '' %}
+    {% set flag = '' | escape %}
     {% set field = milestone.fields.get(tableFieldId) %}
       {% if (field.id == 'date') and (field.value | beforeToDay) %}
-        {% set flag = 'red-date' %}
+        {% set flag = 'red-date' | escape %}
```
gds-gharper, line 188: This is currently a static value, you should escape it when you output, not when you create.

common/macros/critical-milestones-macro.njk
```diff
@@ -28,7 +28,7 @@
             {% for projectField in projectFields %}
               {% set field = project.fields.get(projectField.id) %}
             {% if (field.id == 'uid')%}
-                {% set uid = field.value | escape %}
+                {% set uid = field.value %}
```
gds-gharper: This appears to be used here: `{{ config.paths.projectDetails | replace( ':uid', uid | replace(' ', '%20')) }}` escape will handle this conversion.  If you're only using UID here then perhaps you don't need to replace with %20?

Does this make more sense to move into the javascript file and have a getProjectDetailsUrl function which returns this?

pages/all-data/AllData.html
```diff
@@ -158,16 +158,16 @@ <h3 class="govuk-heading-s panelTitle">Filters</h3>
     {% set field = project.fields.get(tableFieldId) %}
 
     {% if (field.id == 'uid')%}
-    {% set uid = field.value | escape %}
+    {% set uid = field.value %}
```
gds-gharper: You seem to have reverted this change, it should be handled the same way as uid above.  Maybe this should be a helper function in the javascript ?

pages/all-data/AllData.html
```diff
@@ -158,16 +158,16 @@ <h3 class="govuk-heading-s panelTitle">Filters</h3>
...
 
     {% set projectValue = (projectValue.push({html:
 
-    "<a href=" + config.paths.projectDetails | replace(":uid", uid | replace(" ", "%20"))  + " class='govuk-link project-link'>" +
-    "<div class='scrollable'>" + field.value + "</div></a>"
+    "<a href=" + config.paths.projectDetails | replace(":uid", uid | escape | replace(" ", "%20"))  + " class='govuk-link project-link'>" +
+    "<div class='scrollable'>" + field.value | escape + "</div></a>"
       if field.id == 'title'
-        else "<span class=" + cellBlock + field.value + ">" + field.value | default('N/A') + "</span>"
+        else "<span class=" + cellBlock + field.value | escape + ">" + field.value | default('N/A') | escape  + "</span>"
```
gds-gharper, line 170: You probably want
`(cellBlock + field.value) | jsonEscape` and `field.value | default('N/A') | escape ` here.  Also please test these with both values you would expect to be escaped and values that you would expect not to be escaped to make sure they are working as you intend.

common/macros/critical-milestones-macro.njk
AhdabNasir: Fixed this to only escape values coming from the database inside the html block

pages/all-data/AllData.html
```diff
@@ -182,14 +182,14 @@ <h3 class="govuk-heading-s panelTitle">Filters</h3>
...
+    {% set flag = '' %}
     {% set field = milestone.fields.get(tableFieldId) %}
       {% if (field.id == 'date') and (field.value | beforeToDay) %}
-        {% set flag = 'red-date' | escape %}
+        {% set flag = 'red-date' %}
       {% endif %}
 
-      {% set milestoneValue = (milestoneValue.push({html: "<span class=" + flag + ">" + field.value | date("D MMM YYYY") + "</span>" if field.type == 'date' else
-      "<span>" + field.value | default('N/A') + "</span>" }), milestoneValue) %}
+      {% set milestoneValue = (milestoneValue.push({html: "<span class=" + flag + ">" + field.value | escape | date("D MMM YYYY") + "</span>" if field.type == 'date' else
```
gds-gharper, line 191: You probably want to reverse escape and date here

common/macros/critical-milestones-macro.njk
AhdabNasir: Fixed this to only escape values coming from the database inside the html block

pages/project-details/ProjectDetails.html
```diff
@@ -38,31 +38,31 @@
     {% set field = project.fields.get(tableFieldId) %}
 
     {% set projectValue = (projectValue.push({html:
-    "<span class=" + cellBlock + field.value + ">" + field.value | default('N/A') + "</span>" }), projectValue) %}
+    "<span class=" + cellBlock + field.value | escape + ">" + field.value | default('N/A') | escape + "</span>" }), projectValue) %}
```
gds-gharper, line 41: I'd put brackets around `cellBlock + field.value` and use jsonEscape here

pages/all-data/AllData.html
AhdabNasir: Have done this now

pages/all-data/AllData.html
AhdabNasir: Removed this and fixed this to only escape values coming from the database inside the html block

pages/project-details/ProjectDetails.html
```diff
@@ -38,31 +38,31 @@
...
 
       {% if (field.id == 'date') and (field.value | beforeToDay) %}
-        {% set flag = 'red-date' | escape %}
+        {% set flag = 'red-date' %}
         {% set missedMilestonesCount = missedMilestonesCount + 1 %}
       {% endif %}
 
-      {% set milestoneValue = (milestoneValue.push({html: "<span class=" + flag + ">" + field.value | date("D MMM YYYY") + "</span>" if field.id == 'date' else
-      "<span>" + field.value | default('N/A') + "</span>" }), milestoneValue) %}
+      {% set milestoneValue = (milestoneValue.push({html: "<span class=" + flag + ">" + field.value | escape | date("D MMM YYYY") + "</span>" if field.id == 'date' else
```
gds-gharper, line 55: I know in this case you're aware that flag is always safe, but for future developers and defensive behaviour you probably want to jsonEscape flag on this line.  You also probably want to revers escape and date

pages/project-details/ProjectDetails.html
```diff
@@ -38,31 +38,31 @@
...
 
     {% endfor %}
     {% set milestoneFields = (milestoneFields.push(milestoneValue), milestoneFields) %}
 
   {% endfor %}
 
 {% set projectValue = (projectValue.push({
-      html: "<span>" + missedMilestonesCount | escape + "</span>" }, {
-      html: "<span>" + project.fields.get('deliveryTheme').value| default('N/A') + "</span>" }),
+      html: "<span>" + missedMilestonesCount + "</span>" }, {
```
gds-gharper, line 64: There's no harm in pushing this through escape and it means the code is more defensive, even if this is always going to be a number.

common/views/feedback.njk
AhdabNasir: I've tested to make sure what I've done is working by inserting the script tag inside the database. Without my changes it would display the alert but now it is not. Will send you the screenshots on Slack

pages/all-data/AllData.html
AhdabNasir: Removed this escape as it is not coming from the database

pages/all-data/AllData.html
AhdabNasir: In Nunjucks there's no jsonEscape

AhdabNasir: There's no jsonEscape filter in Nunjucks

pages/all-data/AllData.html
AhdabNasir: Fixed this

pages/project-details/ProjectDetails.html
AhdabNasir: There's no jsonEscape in nunjucks

pages/project-details/ProjectDetails.html
AhdabNasir: Reversed data and escape

AhdabNasir: There's no jsonEscape in nunjucks

pages/project-details/ProjectDetails.html
AhdabNasir: Done this

pages/all-data/AllData.html
AhdabNasir: I've escaped the UID inside the html block above

common/macros/critical-milestones-macro.njk
AhdabNasir: It's currently done like so "<a href=" + config.paths.projectDetails | replace(":uid", uid | escape | replace(" ", "%20"))  and using replace twice first to remove the :uid and replace the spacing

pages/all-data/AllData.html
gds-gharper: Ah, more reading was worthwhile for me, looking at the code for escape it'll handle attributes fine I believe

gds-gharper: The other part of the comment that you want to escape after combining is correct though

pages/project-details/ProjectDetails.html
gds-gharper: Yah, as mentioned above escape should work fine

pages/all-data/AllData.html
AhdabNasir: Fixed this

pages/project-details/ProjectDetails.html
AhdabNasir: Done this

pages/project-details/ProjectDetails.html
AhdabNasir: Done on flag

pages/all-data/AllData.html
```diff
@@ -164,10 +164,10 @@ <h3 class="govuk-heading-s panelTitle">Filters</h3>
 
     {% set projectValue = (projectValue.push({html:
 
-    "<a href=" + config.paths.projectDetails | replace(":uid", uid | replace(" ", "%20"))  + " class='govuk-link project-link'>" +
-    "<div class='scrollable'>" + field.value + "</div></a>"
+    "<a href=" + config.paths.projectDetails | replace(":uid", uid | escape | replace(" ", "%20"))  + " class='govuk-link project-link'>" +
```
gds-gharper, line 167: You should be aware of context when you're putting variables into a page.  In this case you're putting in a URL into an attribute.  As such since uid is part of the URL you probably want to URI encode it rather than HTML escape.   This would also take care of the second replace you have here.  As it stands if a UID contained an & or a ? you'd end up with an invalid URL.  There's a filter for this in nunjucks https://mozilla.github.io/nunjucks/templating.html#urlencode

Instead of `replace " ", "%20"` this line should probably be:

```"<a href=" + config.paths.projectDetails | replace(":uid", uid | urlencode)  + " class='govuk-link project-link'>" +```

One of the reasons there are so many edge cases being found here is that you're doing far too much work inside the template.  In general templates should be for display only, any logic should be encapsulated within the actual controller or helper functions or even in the model itself.  It's a little too late to change this now but it's worth bearing in mind for future work.

Quite often when you're writing logic inside a template it's worth thinking "should I encapsulate this within a function" or "am I doing too much work here" (which often amount to the same question).  The two main ways to do that are obviously to write partials, or to write helper functions (there are other options like extending the model or doing the transform within the controller, it depends on the situation).

AhdabNasir: I've done this now and thanks Gareth. I do agree and especially with the All data page, the accordion table was very complex to build. With the other pages, I've now tried to limit logic as much as I can. It's definitely been a learning experience and will take this with me in future projects.

gds-gharper: Did you push your changes here?  I don't see them

AhdabNasir: Sorry just pushed now

PR #134, opened 2020-05-04
fix error occurring when no projects uploaded

dtaylor7:


dtaylor7:
> Could we get a test for this?

all done

PR #136, opened 2020-05-05
Projectlist bug fix

AhdabNasir:


pages/admin/edit-project-field/EditProjectField.html
```diff
@@ -191,12 +191,12 @@ <h1 class="govuk-heading-xl">{{ "Edit project '" + data.displayName + "' field"
     {
       value: "true",
       text: "Yes",
-      checked: true if (data.isActive == "true") else true
+      checked: true if (data.isActive == true) else true
```
dtaylor7, line 194: wont this just always be ticked? maybe should be:
`checked: true if (data.isActive === true)`

AhdabNasir: By default it is ticked on the Add Field page. When the user then goes to edit, it should then reflect whatever the user selected

PR #142, opened 2020-05-07
auto assign ids for projects and milestones if no idea supplied also, increase the timeout to 1 hour

dtaylor7:
also, increase the timeout to 1 hour

### DIFF
--- a/helpers/utils.js
+++ b/helpers/utils.js
@@ -55,17 +55,9 @@ const camelCase = (str = '') => {
 const truthy = [true, 'true', 'yes', 1, '1', 'y'];
 const falsey = ['false', 'no', '0', 'n'];
 
-const pad = n => {
-  if (n < 10) {
-    return "0" + n;
-  }
-  return n;
-};
-
 module.exports = {
   removeNulls,
   truthy,
   falsey,
-  camelCase,
-  pad
+  camelCase
 };
--- a/helpers/utils.js
+++ b/helpers/utils.js
@@ -50,14 +50,22 @@ const camelCase = (str = '') => {
     .replace(/^(.)/, function(b) {
       return b.toLowerCase();
     });
-} ;
+};
 
 const truthy = [true, 'true', 'yes', 1, '1', 'y'];
 const falsey = ['false', 'no', '0', 'n'];
+
+const pad = n => {
+  if (n < 10) {
+    return "0" + n;
+  }
+  return n;
+};
 
 module.exports = {
   removeNulls,
   truthy,
   falsey,
-  camelCase
+  camelCase,
+  pad
 };
--- a/models/milestone.js
+++ b/models/milestone.js
@@ -122,9 +122,10 @@ class Milestone extends Model {
     return fields;
   }
 
-  static async getNextIDIncrement(projectUid, options = {}) {
-    if (options.transaction) {
-      options.lock = options.transaction.LOCK
+  static async getNextIDIncrement(projectUid, { transaction } = {}) {
+    const options = { transaction };
+    if (transaction) {
+      options.lock = transaction.LOCK
     }
 
     const milestone = await Milestone.findOne({
--- a/models/milestone.js
+++ b/models/milestone.js
@@ -5,6 +5,8 @@ const MilestoneFieldEntry = require('./milestoneFieldEntry');
 const MilestoneField = require('./milestoneField');
 const moment = require('moment');
 const pick = require('lodash/pick');
+const { pad } = require('helpers/utils');
+const logger = require('services/logger');
 
 class Milestone extends Model {
   static includes(attributeKey) {
@@ -36,7 +38,6 @@ class Milestone extends Model {
     },{
       name: 'uid',
       type: 'string',
-      isRequired: true,
       isUnique: true,
       importColumnName: 'Milestone UID',
       config,
@@ -76,6 +77,16 @@ class Milestone extends Model {
     const milestoneAttributes = attributes.filter(attribute => this.rawAttributes[attribute]);
     const milestoneToUpsert = pick(milestone, milestoneAttributes);
 
+    if (!milestoneToUpsert.uid) {
+      try {
+        milestoneToUpsert.uid = await Milestone.getNextIDIncrement(milestoneToUpsert.projectUid);
+        milestone.uid = milestoneToUpsert.uid;
+      } catch (error) {
+        logger.error('Unable to generate next milestone uid');
+        throw error;
+      }
+    }
+
     await this.upsert(milestoneToUpsert, options);
 
     await this.importMilestoneFieldEntries(milestone, milestoneFields, options);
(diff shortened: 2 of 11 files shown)


models/project.js
```diff
@@ -87,6 +90,16 @@ class Project extends Model {
...
+    if (!projectToUpsert.uid) {
+      try {
+        projectToUpsert.uid = await Project.getNextIDIncrement(projectToUpsert.departmentName);
+        project.uid = projectToUpsert.uid;
+      } catch (error) {
+        logger.error('Unable to generate next project uid');
+        throw error;
+      }
+    }
+
```
gds-gharper, line 102: So in this it's slightly different to milestones, you should do the following:

1. Create a transaction
2. Fetch the department record, for update with the transaction (using lock: t.LOCK.UPDATE, transaction: t as options to sequelize), this blocks anything else from creating projects within this department
3. Now run your selects which fetch the projects (if any) and calculate the next value (in the same transaction)
4. Insert the new project
5. Commit the transaction

models/project.js
```diff
@@ -121,6 +134,20 @@ class Project extends Model {
...
+      where: { departmentName },
+      order: [['uid', 'DESC']]
+    });
+
+    if(!project) {
+      return `${departmentName}-01`;
+    }
+
+    const current = parseInt(project.uid.split('-')[1]);
+    return `${departmentName}-${pad(current+1)}`;
```
gds-gharper, line 149: You can replace this with `sprintf("%s-%02d",departmentName,current+1)`

models/milestone.js
```diff
@@ -110,6 +121,20 @@ class Milestone extends Model {
...
+      where: { projectUid },
+      order: [['uid', 'DESC']]
+    });
+
+    if(!milestone) {
+      return `${projectUid}-01`;
+    }
+
+    const current = parseInt(milestone.uid.split('-')[2]);
+    return `${projectUid}-${pad(current+1)}`;
```
gds-gharper, line 136: You can replace this with `sprintf("%s-%02d",projectUid,current+1)`

test/unit/helpers/utils.test.js
```diff
@@ -30,4 +30,14 @@ describe('helpers/utils', () => {
...
+
+  describe('#pad', () => {
+    it('returns number padded with 0', () => {
+      expect(utils.pad(4)).to.eql('04');
+    });
+
+    it('returns number not padding with 0', () => {
+      expect(utils.pad(10)).to.eql(10);
+    });
+  })
```
gds-gharper, line 42: I think this should be replaced with sprintf, but I suspect your code fails with negative numbers

dtaylor7: i dont think we really need to handle negative numbers do we?

helpers/utils.js
```diff
@@ -50,14 +50,22 @@ const camelCase = (str = '') => {
...
 const truthy = [true, 'true', 'yes', 1, '1', 'y'];
 const falsey = ['false', 'no', '0', 'n'];
 
+const pad = n => {
+  if (n < 10) {
+    return "0" + n;
+  }
+  return n;
+};
+
```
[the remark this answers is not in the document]
dtaylor7: there is no sprintf for node.

there are other ways of doing this importing additional library, but seems a bit overkill no?

models/milestone.js
```diff
@@ -76,6 +77,16 @@ class Milestone extends Model {
...
+      try {
+        milestoneToUpsert.uid = await Milestone.getNextIDIncrement(milestoneToUpsert.projectUid);
+        milestone.uid = milestoneToUpsert.uid;
+      } catch (error) {
+        logger.error('Unable to generate next milestone uid');
+        throw error;
+      }
+    }
+
     await this.upsert(milestoneToUpsert, options);
```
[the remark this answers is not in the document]
dtaylor7: what does select for update do?

helpers/utils.js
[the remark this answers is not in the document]
gds-gharper: https://www.npmjs.com/package/sprintf-js would handle it, and whilst we don't need to handle negative numbers right now it means we can handle other formats trivially and we're not implementing functionality which is almost certainly better tested and proven than anything we'd write ourselves.

models/milestone.js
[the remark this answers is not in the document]
gds-gharper: It selects a row (or rows) but marks them as being for update, which locks them in the transaction and prevents other processes from updating them or selecting them for update themselves.  Preventing any race conditions.

models/milestone.js
```diff
@@ -76,6 +77,16 @@ class Milestone extends Model {
     const milestoneAttributes = attributes.filter(attribute => this.rawAttributes[attribute]);
     const milestoneToUpsert = pick(milestone, milestoneAttributes);
 
+    if (!milestoneToUpsert.uid) {
+      try {
+        milestoneToUpsert.uid = await Milestone.getNextIDIncrement(milestoneToUpsert.projectUid, options);
```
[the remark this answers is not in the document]
dtaylor7: i am using transaction for the entire import, that way if any of it fails it will role back. if you would like that to change, then would need a separate task to change that functionality as it increases the scope of this functionality

i have tested will multiple milestones and projects and it is working correctly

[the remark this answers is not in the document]
gds-gharper: Ok, that seems sensible, I'd be tempted to explicitly pass in the transaction rather than "options" as it makes the functions much less black boxy.  At the very least the lock should only be performed on the SELECT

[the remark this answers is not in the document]
dtaylor7: 👍

models/milestone.js
```diff
@@ -110,6 +121,24 @@ class Milestone extends Model {
 
     return fields;
   }
+
+  static async getNextIDIncrement(projectUid, options = {}) {
+    if (options.transaction) {
+      options.lock = options.transaction.LOCK
```
gds-gharper, line 127: This will modify the callers object which you probably don't want to do, you should clone the object if you need to modify it since javascript passes objects by reference.  You also want to pass in `lock: options.transaction.LOCK.UPDATE` since that's the type of lock you want (I'm not sure what happens if you just pass in the LOCK object).

Have you tested to make sure the SQL generated includes FOR UPDATE?

Same issues with the project fields below

dtaylor7: can you just write it how you would like it?

dtaylor7: that any better?

models/milestone.js
```diff
@@ -110,6 +121,25 @@ class Milestone extends Model {
 
     return fields;
   }
+
+  static async getNextIDIncrement(projectUid, { transaction } = {}) {
+    const options = { transaction };
+    if (transaction) {
+      options.lock = transaction.LOCK
```
gds-gharper, line 128: I've tested and passing in the entire enum here does appear to work

PR #145, opened 2020-05-08
add milestone details page and remove missed milestones count from al…

dtaylor7:
…l-data table

### DIFF
--- a/models/milestone.js
+++ b/models/milestone.js
@@ -46,12 +46,14 @@ class Milestone extends Model {
       name: 'description',
       type: 'string',
       importColumnName: 'Milestone description',
+      displayName: 'Description',
       config,
       description: 'Name and short description of the milestone, covering what it is, who owns it, what form it takes and why it is required.'
     },{
       name: 'date',
       type: 'date',
       importColumnName: 'Target date for delivery',
+      displayName: 'Target date for delivery',
       config: {
         exportOptions: {
           header: { fill: "6C9EE9" },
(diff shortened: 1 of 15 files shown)


models/milestone.js
```diff
@@ -135,6 +137,13 @@ Milestone.init({
       if (!this.getDataValue('date')) return;
       return moment(this.getDataValue('date'), 'YYYY-MM-DD').format('DD/MM/YYYY');
     }
+  },
+  updatedAt: {
+    type: DATE,
```
gds-gharper, line 142: This should probably be a timestamp, not a date, how does this interact with the updatedAt field below?

models/milestoneFieldEntryAudit.js
```diff
@@ -21,7 +22,10 @@ MilestoneFieldEntryAudit.init({
   archivedAt: {
     type: DATE,
```
gds-gharper, line 23: This should probably be a TIMESTAMP, I know it's not something we changed here though

models/milestone.js
dtaylor7: its just for reading the value from the database

dtaylor7: this is a timestamp, see docs: https://sequelize.org/v5/manual/data-types.html

models/milestoneFieldEntryAudit.js
dtaylor7: this is a timestamp, see docs: https://sequelize.org/v5/manual/data-types.html

PR #147, opened 2020-05-11
fix exporting excel, also format dates correctly

dtaylor7:


dtaylor7:
> It'd be good to get a test for that, but I'll approve for now, can you add a story for it.

will do now

PR #163, opened 2020-05-14
Upcoming milestones view

AhdabNasir:


common/javascript/see-more.js
```diff
@@ -0,0 +1,21 @@
...
+
+  function changeHandler() {
+    let information = this.parentNode.nextElementSibling;
+
+    if (this.innerHTML == 'See more') {
+      information.style.display = 'block';
+      this.innerHTML = 'See less';
+    } else if (this.innerHTML == 'See less') {
+      information.style.display = 'none';
+      this.innerHTML = 'See more';
```
gds-gharper, line 13: These days you would often add/remove a class to the tag rather than change the inner style.  That way you can check for that class (or use data attributes) to see whether you should expand or collapse.

pages/all-data/AllData.js
```diff
@@ -103,8 +103,17 @@ class AllData extends Page {
   }
 
   async postRequest(req, res) {
+    let filters = this.data.filters;
+
+    if(req.body.clear) {
+      for (var key in filters) {
+        delete filters[key];
+      }
+      this.saveData(removeNulls({ filters }));
```
gds-gharper, line 112: Is this required since you just deleted everything in the hash?

Could you replace this whole block with `this.saveData({})`

pages/upcoming-milestones/UpcomingMilestones.html
```diff
@@ -3,78 +3,68 @@
...
-
-  <a class="govuk-link" href="{{ config.paths.impactDefinitions }}">View impact and confidence definitions</a>
-  <br><br><br><br>
-
-  {{ govukAccordion({
-    id: "accordion-table",
-    classes: "upcoming-milestones-accordion",
-    items: upcomingMilestonesTable
-  }) }}
+<div class="govuk-grid-row" id="show">
```
gds-gharper, line 18: I'm not 100% sure "show" is a good name for an id (which is supposed to be unique)

pages/upcoming-milestones/UpcomingMilestones.js
```diff
@@ -1,81 +1,179 @@
...
 
+  get schema() {
+    return {
+      filters: {
+        date: {
+          from: moment().format('DD/MM/YYYY'),
+          to: moment().add('6', 'weeks').format('DD/MM/YYYY')
+        }
+      }
+    };
```
gds-gharper, line 21: One of the problems with this is something I've been thinking about for a while.  With all of these filters we're not currently indexing on value.  I should probably add an index to the ..._field tables on ..._field_id and value

That may be enough to improve the performance here

pages/upcoming-milestones/UpcomingMilestones.js
```diff
@@ -1,81 +1,179 @@
...
+
+    return this.groupDataByDate(departments);
+  }
+
+  projectsWithMilestonesForDate(projects, date) {
+    const projectsWithMilestonesOnDate = projects.map(project => {
+      // only milestones on this day
+      project.milestones = project.milestones.filter(milestone => milestone.date === date.format('DD/MM/YYYY'));
+      return project;
     });
```
gds-gharper, line 35: We'll leave it for now but this could easily be handled by the DAO with either no, or very minimal changes.  I won't let it get in the way of this PR but will look at it tomorrow.

pages/upcoming-milestones/UpcomingMilestones.js
```diff
@@ -1,81 +1,179 @@
...
+  async postRequest(req, res) {
+    let filters = this.data.filters;
+
+    if(req.body.clear) {
+      for (var key in filters) {
+        delete filters[key];
+      }
+      this.saveData(removeNulls({ filters }));
+      return res.redirect(this.url);
+    }
```
gds-gharper, line 157: This looks exactly the same as the one in the AllData.js page, maybe push it into a single helper function?

pages/upcoming-milestones/UpcomingMilestones.js
```diff
@@ -1,81 +1,179 @@
...
+      let filters = this.data.filters;
+      const filterId = req.body.filterId;
+      const optionValue = req.body.optionValue;
+
+      if (!optionValue) {
+        delete filters[filterId];
+      } else {
+        for (var i = filters[filterId].length - 1; i >= 0; i--) {
+          if (filters[filterId][i] == optionValue) filters[filterId].splice(i, 1);
+        }
```
gds-gharper, line 169: Minor, but I assume multiple filters can't match the same filterId, so you can break when you hit the if statement.  It's not a big deal though since I doubt there'll every be more than a few entries in that array.

pages/upcoming-milestones/partials/filters/dates.njk
```diff
@@ -0,0 +1,24 @@
+{{ govukInput({
+  label: {
+    text: "From"
+  },
+  classes: "govuk-input--width-10",
+  hint: {
+    html: "For example, 01/01/2020"
```
gds-gharper, line 7: There's no formatting in here, why not simply use the `text:` option instead of html.  It's a lot safer.

PR #164, opened 2020-05-15
Async filter missing

gds-gharper:


services/dao.js
```diff
@@ -163,17 +163,17 @@ class DAO {
         LEFT OUTER JOIN project_field AS \`projectFieldEntries->projectField\` ON (\`projectFieldEntries->projectField\`.id = \`projectFieldEntries\`.project_field_id)
     `;
 
-    projectFieldFilters.forEach( async(filter) => {
+    await projectFieldFilters.forEach( async(filter) => {
```
dtaylor7, line 166: you cannot await on `foreach`, you will need to convert to `for in`

services/dao.js
```diff
@@ -163,17 +163,17 @@ class DAO {
         LEFT OUTER JOIN project_field AS \`projectFieldEntries->projectField\` ON (\`projectFieldEntries->projectField\`.id = \`projectFieldEntries\`.project_field_id)
     `;
 
-    projectFieldFilters.forEach( async(filter) => {
+    for await (const filter of projectFieldFilters) {
```
dtaylor7, line 166: does this work? it should look like this:

```
for (const filter of projectFieldFilters) {
      const fieldId = await dao.getProjectFieldId(filter);
      query += this.generateProjectProjectFieldFilter(fieldId,filter.value);
}
```

dtaylor7: no `await` after for

PR #173, opened 2020-05-18
Missed milestone filters milestone status

dtaylor7:
- add extra milestone specific filters to upcoming milestones page
- add milestone status to all-data and project-details milestone table

### DIFF
--- a/helpers/filters.js
+++ b/helpers/filters.js
@@ -92,18 +92,18 @@ const getProjectCoreFields = async (search, user) => {
 const getProjectFields = async (search, user) => {
   const response = await ProjectField.findAll({
     attributes: [
-      [literal(`projectField.id`), 'id'],
-      [literal(`projectField.name`), 'name'],
-      [literal(`projectField.display_name`), 'displayName'],
-      [literal(`projectField.type`), 'type'],
-      [literal(`projectFieldEntries.value`), 'value']
+      [literal('projectField.id'), 'id'],
+      [literal('projectField.name'), 'name'],
+      [literal('projectField.display_name'), 'displayName'],
+      [literal('projectField.type'), 'type'],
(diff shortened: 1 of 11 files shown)


helpers/filters.js
```diff
@@ -166,13 +168,100 @@ const getProjectFields = async (search, user) => {
...
-const getFilters = async (search = {}, user) => {
-  const filters = await getProjectCoreFields(search, user);
+const getMilestoneFields = async (search, user) => {
+  const response = await MilestoneField.findAll({
+    attributes: [
+      [literal(`milestoneField.id`), 'id'],
+      [literal(`milestoneField.name`), 'name'],
+      [literal(`milestoneField.display_name`), 'displayName'],
+      [literal(`milestoneField.type`), 'type'],
+      [literal(`milestoneFieldEntries.value`), 'value']
```
gds-gharper, line 178: Why the backticks here?

helpers/filters.js
```diff
@@ -166,13 +168,100 @@ const getProjectFields = async (search, user) => {
...
+      });
+    } else {
+      existingField.options.push({
+        value
+      });
+      existingField.options.sort((a, b) => (a.value > b.value) ? 1 : -1);
+    }
+
+    return fields;
+  }, new Map());
```
gds-gharper, line 252: Any particular reason to use a Map here over a normal object?

dtaylor7: means you can iterate over

helpers/filters.js
dtaylor7: no reason probably had dynamic at some point

helpers/filters.js
dtaylor7: was originally an array, so code had been written to iterate over. However we needed the ability to find specific items by name, so rather than change existing code I changed to map.

PR #183, opened 2020-05-19
Add edit project page

dtaylor7:
- Added edit project page
- Added new url
- Add new link on project detail page


pages/edit-project/EditProject.js
```diff
@@ -0,0 +1,174 @@
...
+      if(project.fields.get(field.name)) {
+        data[field.name] = project.fields.get(field.name).value;
+      }
+      return data;
+    }, {});
+
+    this.saveData(data);
+  }
+
+  transformDeliveryConfidenceValue(value) {
```
gds-gharper, line 144: Would this logic also be in the spreadsheet uploader?  Is it worth combining them into one block?

pages/edit-project/partials/edit.njk
```diff
@@ -0,0 +1,124 @@
...
+          href: "#"
+        }]
+    }) }}
+  {% endif %}
+
+<form action='{{ editModeUrl }}' method='POST'>
+
+  {% for projectGroup in projectEntryGroups %}
+    <br>
+    <h2 class="govuk-heading-m">{{ projectGroup.name }}</h2>
```
gds-gharper, line 28: It's probably worth displaying the UID here too

pages/edit-project/partials/edit.njk
```diff
@@ -0,0 +1,124 @@
...
+          },
+          hint: {
+            text: field.description
+          },
+          value: data[field.name]
+        }) }}
+
+        {% elseif (field.type == 'integer' or field.type == 'date')  %}
+
+        {% set value = data[field.name] | date("DD/MM/YYYY") if field.type == 'date' else data[field.name] %}
```
gds-gharper, line 52: I'm assuming these are the same for milestones, maybe make them global partials which take the field record as a parameter?  Have one for each field type

pages/edit-project/EditProject.js
dtaylor7: Currently, it's only used milestone details page and edit page, it could be shared, however.

AhdabNasir:
> It would be useful to have some tests for this, but I'm aware that time is of the essence and this is a useful tool for the admins which will reduce reliance on developers to edit data.

uNIT=

pages/edit-project/EditProject.js
```diff
@@ -0,0 +1,164 @@
...
+  get signOffMode() {
+    return this.req.params && !this.req.params.edit && !this.req.params.successful;
+  }
+
+  get editMode() {
+    return this.req.params && this.req.params.edit;
+  }
+
+  get successfulModeUrl() {
+    return `${paths.editProject}/${this.req.params.uid}/successful`
```
gds-gharper, line 36: What happens if someone sends a maliciously crafted uid to this?  You may wish to validate that it only contains valid uid characters, or at the very least make sure it doesn't contain / or .. or a few other things

pages/edit-project/partials/edit.njk
gds-gharper: We still don't appear to be showing the UID?

pages/edit-project/EditProject.js
AhdabNasir: I've added the UID as a caption in the heading top of the page so the user knows what they are editing.

dtaylor7: @gds-gharper we dont handle this type of security, on milestone edit page either, can we add this as separate task in backlog?

gds-gharper: can do, please add a task

dtaylor7: https://technologyprogramme.atlassian.net/browse/DDB-633

PR #189, opened 2020-07-14
DDB-463 - change password on reset

dtaylor7:
Update the user model to reference the correct column name in database


scripts/resetUser.js
```diff
@@ -23,6 +23,7 @@ if (!email) {
 let passphrase = randomString.generate({
   readable: true
 });
+console.log(passphrase)
```
gds-gharper, line 26: Remove the debugging please.  For future reference if there's no notify api key in the config the script will output the password anyway

PR #193, opened 2020-07-15
add tableau service to get iframe url

dtaylor7:
Example of how to use inside page:
```
const tableau = require('services/tableau');
....
async getIframeUrl() {
  return await tableau.getTableauUrl(this.req.user, 'workbook', 'view');
}
```

in HTML: 
```
<iframe src="{{ getIframeUrl() }}></iframe>
```

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -11,8 +11,7 @@
     },
     "googleAnalytics": {},
     "tableau": {
-      "url": "someurl",
-      "username": "someusername"
+      "url": "someurl"
     }
   },
   "bcrypt": {
--- a/config/default.json
+++ b/config/default.json
@@ -10,9 +10,7 @@
       "name": "Transition Taskforce Dashboard"
     },
     "googleAnalytics": {},
-    "tableau": {
-      "url": "someurl"
-    }
+    "tableau": {}
   },
   "bcrypt": {
     "saltRounds": 10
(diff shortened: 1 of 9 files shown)


config/default.json
```diff
@@ -9,7 +9,10 @@
     "tfa": {
       "name": "Transition Taskforce Dashboard"
     },
-    "googleAnalytics": {}
+    "googleAnalytics": {},
+    "tableau": {
+      "url": "someurl"
```
gds-gharper, line 14: I wouldn't default this, especially to an invalid value

config/production.json
```diff
@@ -13,6 +13,9 @@
     },
     "googleAnalytics": {
       "id": "UA-161904342-1"
+    },
+    "tableau": {
+      "url": ""
```
gds-gharper, line 18: If you really wanted to make sure the key was known, I'd set it to null rather than an empty string

services/tableau.js
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  const response = await fetch(`${config.services.tableau.url}/trusted`, {
+    method: 'POST',
+    body: {
+      username: user.email
+    }
+  });
+
+  return `http://${config.services.tableau.url}/trusted/${response}/views/${workbook}/${view}`;
```
gds-gharper, line 16: Should we check the result of the fetch before merging it into the string?  If it fails then we could be creating a random URL.

dtaylor7: if the request fails then an error will be thrown and it wont reach here

PR #194, opened 2020-07-15
404-page bug fix

AhdabNasir:
When a user types in an address that doesn't exist, it will redirect to the 404 page

### DIFF
--- a/app.js
+++ b/app.js
@@ -32,3 +32,7 @@ pages.attach(app);
 app.get('/', (req, res) => res.redirect(startPage()));
 
 logger.attachErrorLogger(app);
+
+app.use(function (req, res) {
+  res.status(404).redirect(config.paths.pageNotFound);
+});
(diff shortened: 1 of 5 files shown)


app.js
```diff
@@ -30,5 +30,7 @@ logger.attachRouteLogger(app);
 cache.preventCache(app);
 pages.attach(app);
 app.get('/', (req, res) => res.redirect(startPage()));
+app.use((req, res) => res.status(404).redirect(config.paths.pageNotFound));
```
gds-gharper, line 33: I don't think we should actually redirect the user to a different page/url, simply show a nicer error page at the url they are currently attempting to load.

Redirecting them implies to the browser that the page they were trying to load was actually somewhere else.  It would also mean that instead of getting an error response they'd get a 302 (and then a 200 on the redirected URL).

AhdabNasir: @dtaylor7 Do you have an idea on how we could do this? Rather than redirect, we just show the 404 page I built?

dtaylor7: its a bit of pain but could do:
```
const 404Page = require('pathj to page')
const p = new 404Page('pathj to page');
app.use(p.router);
```

dtaylor7: this is now resolved @gds-gharper

PR #197, opened 2020-07-16
reporting overview

AhdabNasir:


pages/reporting-overview/ReportingOverview.html
```diff
@@ -0,0 +1,25 @@
...
+
+{% block beforeContent %}
+{{ govukBackLink({
+  text: "Back",
+  href: "#",
+  attributes: {
+    onclick: "history.go(-1);"
+  }
+}) }}
+{% endblock %}
```
gds-gharper, line 12: It's fine for now since we do it in other places as well.  But the back link is breaking some other areas of the site.  I'd rather replace this with a breadcrumb and let them choose where to go.  They can always use the back button on their browser to go back.  There should be a story in the backlog to change this.

Worth bearing in mind for future.

AhdabNasir: I've realised this too! Theres a story in the backlog for this so hopefully will fix soon.

PR #198, opened 2020-07-16
Notification for upcoming milestones

dtaylor7:


config/default.json
```diff
@@ -1,6 +1,7 @@
 {
   "env": "development",
   "port": "3000",
+  "serviceUrl": "http://localhost:3000",
```
gds-gharper, line 4: It's not a great idea to default things which would break badly when they go to live.

notifications/upcomingMilestones.js
```diff
@@ -0,0 +1,115 @@
...
+      } catch (error) {
+        logger.error(`Error sending upcoming milestones email to ${user.email}". ${error}`);
+      }
+    }
+  }
+};
+
+const setLock = async () => {
+  const guid = uuidv4();
+  const query = `UPDATE dashboard_locks SET guid="${guid}" WHERE name='upcoming milestones notifications'`;
```
gds-gharper, line 86: `and guid IS NULL`, otherwise two running at the same time will replace each other.  I would also not use string joining to create the SQL, let sequelize handle string escaping and quoting (in this case I know the guid will almost certainly never have bad characters in, but it's a bad habit to get into).

```
await sequelize.query(
  "UPDATE dashboard_locks SET guid=? WHERE name=?",
  { replacements: [guid,"upcoming milestones notifications] }
);
```

Also, I would suggest putting the name of the lock into config somewhere, so we're not using magic strings multiple times in the code.

notifications/upcomingMilestones.js
```diff
@@ -0,0 +1,115 @@
...
+  await sequelize.query(query);
+  return guid;
+};
+
+const getLock = async (guid) => {
+  const query = `SELECT * FROM dashboard_locks WHERE guid="${guid}" AND name='upcoming milestones notifications'`;
```
gds-gharper, line 92: Use sequelize to quote and escape the parameters.

notifications/upcomingMilestones.js
```diff
@@ -0,0 +1,115 @@
...
+  return response[0].length > 0;
+};
+
+const notifyUpcomingMilestones = async () => {
+  const guid = await setLock();
+
+  if(await getLock(guid)) {
+    const projects = await getProjectsDueInNextTwoDays();
+    const projectsByDepartment = groupProjectsByDepartment(projects);
+    await sendEmails(projectsByDepartment);
```
gds-gharper, line 103: Don't forget to remove the lock after finishing (set guid to NULL).  You almost certainly want to do this in a catch/finally as well to make sure it's always unset.  It's probably worthwhile sending us (the transition team) a notification when the process completes to summarise what was done and any errors (you could just log it but we're unlikely to watch the logs, so in this case a notification is better).

dtaylor7: this would happen every day? how best to send the notification?

gds-gharper: Yes, it'll happen every day.  I'd be tempted just to send an email.

dtaylor7: to which email address?

gds-gharper: Stick it in config, but you can use mine for now.  We can try and get a group/mailing list set up after.

dtaylor7: is there a way of sending emails from the logs? for example, papertrail sends an email every X hours when an error Is logged?

gds-gharper: Not currently, just set up a template in notify and use that.  We just need a brief summary to let people know it ran and how many emails it sent out I guess (and any errors).  If we need to dig further we can go to the logs

config/default.json
```diff
@@ -60,6 +60,12 @@
     "missedMilestones": false
   },
   "notify": {
-    "createTemplateKey": "ca1dcdac-7d98-4158-be9d-5190ceb62289"
+    "createTemplateKey": "ca1dcdac-7d98-4158-be9d-5190ceb62289",
+    "missedMilestonesKey": "81ac6df8-2e4a-4d33-9ba9-7d6d9bca46fb",
+    "summaryNotificationKey": "26be1793-7401-4541-b119-78ce1d6a729b",
+    "summaryNotificationEmail": "gareth.harper@digital.cabinet-office.gov.uk"
```
gds-gharper, line 66: I thought we were putting this in environment variables so we didn't send peoples email addresses into the public repo?

notifications/upcomingMilestones.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  return summary;
+};
+
+const setLock = async () => {
+  const guid = uuidv4();
+  const query = `UPDATE dashboard_locks SET guid=? WHERE name=? AND guid IS NULL`;
+  const options = {
+    replacements: [ guid, config.locks.upcomingMilestonesCotifications ]
```
gds-gharper, line 93: upcomingMilestonesCotifications should probably be upcomingMilestonesNotifications everywhere

config/default.json
dtaylor7: yes, we had that conversation after i had done this work

config/default.json
```diff
@@ -60,6 +60,12 @@
     "missedMilestones": false
   },
   "notify": {
-    "createTemplateKey": "ca1dcdac-7d98-4158-be9d-5190ceb62289"
+    "createTemplateKey": "ca1dcdac-7d98-4158-be9d-5190ceb62289",
+    "missedMilestonesKey": "81ac6df8-2e4a-4d33-9ba9-7d6d9bca46fb",
+    "summaryNotificationKey": "26be1793-7401-4541-b119-78ce1d6a729b",
+    "summaryNotificationEmail": ""
```
gds-gharper, line 66: probably better to be null not an empty string

migrations/10_notifications_lock.js
```diff
@@ -0,0 +1,36 @@
...
+const Sequelize = require('sequelize')
+
+const up = async (query) => {
+  await query.createTable('dashboard_locks', {
+    guid: Sequelize.DataTypes.STRING(50),
+    name: Sequelize.DataTypes.STRING(50)
+  }, { charset: services.mysql.charset });
+
+  query.bulkInsert('dashboard_locks', [{
+    name: 'upcoming milestones notifications'
```
gds-gharper, line 12: Might as well use the value from the config for this

config/default.json
dtaylor7: you cant have NULL in json

PR #216, opened 2020-08-03
Add new entity tables and import entity functionality 

dtaylor7:
- add new table migrations
 - add new table models
 - add new pages to be able to import data
 - sample data


migrations/11_entity_tables.js
```diff
@@ -0,0 +1,333 @@
...
+
+const up = async (query) => {
+  await query.createTable('category', {
+    id: {
+      type: Sequelize.DataTypes.INTEGER,
+      primaryKey: true,
+      autoIncrement: true
+    },
+    name: {
+      type: Sequelize.DataTypes.STRING(50),
```
gds-gharper, line 15: Make sure this is unique

models/categoryField.js
```diff
@@ -0,0 +1,46 @@
...
+  categoryId: {
+    field: 'category_id',
+    type: INTEGER
+  },
+  name: STRING(45),
+  displayName: {
+    field: "display_name",
+    type: STRING(300),
+  },
+  type: ENUM("string", "boolean", "integer", "float", "group"),
```
gds-gharper, line 21: date/timestamp?

models/entityFieldEntryAudit.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+  value: {
+    type: TEXT
+  },
+  archivedAt: {
+    type: DATE,
+    field: 'archived_at',
+    defaultValue: NOW
+  }
+}, { sequelize, modelName: 'entityFieldEntryAudit', tableName: 'entity_field_entry_audit', createdAt: 'created_at', updatedAt: 'updated_at' });
```
gds-gharper, line 23: I'm not sure we want sequelize to automatically update the created_at and updated_at fields, we want them to be the same as they were in the entity_field_entry table, probably best to make sure the createdAt and updatedAt fields don't change when a row is created.

pages/data-entry-entity/entity-bulk-upload/partials/upload-file.njk
```diff
@@ -0,0 +1,51 @@
...
+<div class="govuk-grid-column-two-thirds">
+
+  <h1 class="govuk-heading-xl">Upload your file</h1>
+
+  {% if flash %}
+    {{ govukErrorSummary({
+      titleText: "There is a problem",
+      errorList: [
+        {
+          html: flash,
```
gds-gharper, line 18: Best to make sure the flash is escaped somewhere

pages/data-entry-entity/entity-import/EntityImport.html
```diff
@@ -0,0 +1,180 @@
...
+  preventDoubleClick: true
+}) }}
+
+</form>
+
+{{ govukDetails({
+  summaryText: "The data uploaded is not what I was expecting",
+  html: "<p class='govuk-body'>If the number of items uploaded don't correspond to the number in the
+  file you uploaded, please log the issue here.</p>
+  <a class='govuk-link' href='mailto:ttf-portfoliodashboard@cabinetoffice.gov.uk'>ttf-portfoliodashboard@cabinetoffice.gov.uk</a>"
```
gds-gharper, line 145: Should the email here come from config, so it can be different in different environments?

pages/data-entry-entity/entity-import/EntityImport.js
```diff
@@ -0,0 +1,187 @@
...
+    };
+  }
+
+  async finaliseImport(importId) {
+    const activeImport = await BulkImport.findOne({
+      where: {
+        userId: this.req.user.id,
+        id: importId,
+        category: {
+          [Op.not]: 'data-entry-old'
```
gds-gharper, line 103: Should we just use NULL instead of a magic string?

pages/data-entry-entity/entity-output/EntityOutput.html
```diff
@@ -0,0 +1,38 @@
...
+      &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-> {{ entityChild.publicId }}<br>
+      {% for entityChildChild in entityChild.children %}
+        &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;--> {{ entityChildChild.publicId }}<br>
+        {% for entityChildChildChild in entityChildChild.children %}
+          &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;--> {{ entityChildChildChild.publicId }}<br>
+        {% endfor %}
+      {% endfor %}
+    {% endfor %}
+    <br><br><br>
+  {% endfor %}
```
gds-gharper, line 35: I'm aware this is just temporary, but could just use nested `<ul></ul>` which should give you the nesting you want as well.  You can even change the bullet to something like an arrow (possibly from fontawesome if you can't find one directly in html)

pages/data-entry-entity/entity-output/EntityOutput.js
```diff
@@ -0,0 +1,41 @@
...
+
+  get middleware() {
+    return [
+      ...authentication.protect(['administrator'])
+    ];
+  }
+
+  async output() {
+    return await Entity.findAll({
+      where: { categoryId: 1 },
```
gds-gharper, line 24: instead of finding the first category by ID, you could search for categories without a parent.  Not sure if we should just loop ourselves and feed the result ids into the next loop instead of nesting an arbitrary 3 levels

helpers/parse.js
```diff
@@ -39,7 +39,7 @@ const parseValue = (value, definition = {}) => {
 const parseItems = (items, itemDefinitions) => {
   const parseItem = item => {
     const parsed = itemDefinitions.reduce((parsedItem, itemDefinition) => {
-      const value = item[itemDefinition.importColumnName];
+      const value = item[itemDefinition.importColumnName || itemDefinition.displayName];
```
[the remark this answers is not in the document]
dtaylor7: no, as projects and milestones also have a displayName, reversing it would break old import

migrations/11_entity_tables.js
```diff
@@ -0,0 +1,333 @@
...
+        model: 'category',
+        key: 'id'
+      },
+      onUpdate: 'cascade',
+      onDelete: 'cascade'
+    }
+  }, { charset: services.mysql.charset });
+
+  query.bulkInsert('category_parent', [{
+    category_id: 2,
```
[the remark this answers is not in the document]
dtaylor7: this will only ever happen when table is empty

migrations/11_entity_tables.js
```diff
@@ -0,0 +1,333 @@
...
+    }
+  }, { charset: services.mysql.charset });
+
+  await query.createTable('entity_parent', {
+    entity_id: {
+      type: Sequelize.DataTypes.INTEGER,
+      references: {
+        model: 'entity',
+        key: 'id'
+      },
```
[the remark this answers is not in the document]
dtaylor7: yes this is working, workbench showing correct relationships

models/entityFieldEntryAudit.js
dtaylor7: this works correctly as we parse in created_at and updated_at so sequelize ignores them

pages/data-entry-entity/entity-import/EntityImport.html
dtaylor7: its just a link, same as text. i don't think there is any need user on staging to looking up who to email here?

pages/data-entry-entity/entity-import/EntityImport.js
dtaylor7: we could do, but in a few weeks we can remove this when we migrate projects and milestones over?

PR #220, opened 2020-08-06
remove staging specific cron

dtaylor7:


dtaylor7:
this is now working, so reverting back to 1 email a day

PR #226, opened 2020-08-17
Entity field management

dtaylor7:


models/category.js
```diff
@@ -54,7 +54,8 @@ class Category extends Model {
         required: true
       },
       where: { isActive: true },
-      raw: true
+      raw: true,
+      order: [['order', 'ASC']]
```
gds-gharper, line 58: please don't create fields which map directly to SQL keywords (like order), priority might work better

pages/admin/category-field-list/CategoryFieldList.js
```diff
@@ -0,0 +1,73 @@
...
+        const values = { order: field.order };
+        await CategoryField.update(values, {
+          where: { id: field.id },
+          transaction
+        });
+      }
+
+      await transaction.commit();
+    } catch (error) {
+      this.req.flash(`Error saing field order: ${error}`);
```
gds-gharper, line 59: saing => saving

pages/admin/edit-category-field/EditCategoryField.html
```diff
@@ -0,0 +1,244 @@
...
+    errorList: [{
+        text: flash,
+        href: "#field-area"
+    }]
+}) }}
+{% endif %}
+
+<form action='{{ req.originalUrl }}' method='POST'>
+
+<input type="hidden" name="id" value="{{ data.id | default('temp') }}" />
```
gds-gharper, line 71: Feels like a magic string, could simply not put the field in if there's no ID and check for the absence of the field to create.  Not massively important though

pages/admin/edit-category-field/EditCategoryField.html
```diff
@@ -0,0 +1,244 @@
...
+      text: "Yes / No",
+      value: "boolean",
+      selected: true if (data.type == 'boolean')
+    }
+  ]
+}) }}
+
+<div id="dropDownOptions">
+  <label class="govuk-label" for="drop-down-options">Drop down options
+  <a href="#/" id="add-option" class="govuk-link">Add another</a></label>
```
gds-gharper, line 139: Why the / after the anchor?

pages/admin/edit-category-field/EditCategoryField.html
```diff
@@ -0,0 +1,244 @@
...
+  {% if data.config and data.type == 'group' %}
+  {% for groupItem in data.config.options %}
+      <input class="govuk-input drop-down-option" value="{{ groupItem }}" name="config[options][]" type="text">
+      <a href="#/" class="govuk-link delete-option">Delete</a>
+      <a href="#/" class="govuk-link disable-option">Disable</a>
+  {% endfor %}
```
gds-gharper, line 145: Probably should indent this once

pages/admin/edit-category-field/EditCategoryField.html
```diff
@@ -0,0 +1,244 @@
...
+    legend: {
+      text: "Is this field active?",
+      classes: "govuk-label"
+    }
+  },
+  items: [
+    {
+      value: "true",
+      text: "Yes",
+      checked: true if (data.isActive == true) else true
```
gds-gharper, line 168: You may wish to check this line, doesn't it always evaluate to true?

Might be worth simply setting it to `(data.isActive == true)`

pages/admin/edit-category-field/EditCategoryField.html
```diff
@@ -0,0 +1,244 @@
...
+    legend: {
+      text: "Is this field mandatory?",
+      classes: "govuk-label"
+    }
+  },
+  items: [
+    {
+      value: "1",
+      text: "Yes",
+      checked: true if (data.isRequired == true) else true
```
gds-gharper, line 192: Same problem here with truth

pages/admin/edit-category-field/EditCategoryField.html
```diff
@@ -0,0 +1,244 @@
...
+  name: "isActive",
+  fieldset: {
+    legend: {
+      text: "Is this field active?",
+      classes: "govuk-label"
+    }
+  },
+  items: [
+    {
+      value: "true",
```
gds-gharper, line 166: We're using the strings "true" and "false" here, but below we're using 1 and 0, any particular reason, they're both denoting true/false.  Personally I think 1/0 maps a little simpler

pages/admin/edit-category-field/EditCategoryField.js
```diff
@@ -0,0 +1,161 @@
...
+    return this.res.redirect(`${url}/summary`);
+  }
+
+  async saveFieldToDatabase() {
+    const field = this.data;
+
+    if (field.id === 'temp') {
+      delete field.id;
+
+      const fields = await Category.fieldDefintions(this.res.locals.category.name);
```
gds-gharper, line 91: fieldDefintions => fieldDefinitions

models/category.js
dtaylor7: it doesnt matter as they are wrapped at query time `?

pages/admin/edit-category-field/EditCategoryField.html
dtaylor7: am using temp to check if i need to create or update

models/category.js
gds-gharper: It's really very bad practice and just asking for trouble down the line.  We have some instances of this and I'd rather get them fixed (it's in the backlog) I'd much rather it doesn't get worse and it's simpler to change it now than later.  It's the equivalent of creating a function called 'require' or 'import' and makes things much harder to read when you're checking the SQL

pages/admin/edit-category-field/EditCategoryField.html
dtaylor7: I am not sure, I will ask @AhdabNasir as this was copied from her work editing fields in milestones

models/category.js
dtaylor7: what shall i change it too?

gds-gharper: priority was a suggestion

dtaylor7: ok

models/category.js
```diff
@@ -48,15 +48,22 @@ class Category extends Model {
...
         attributes: [],
         model: Category,
         where: { name: categoryName },
         required: true
       },
-      where: { isActive: true },
-      raw: true
+      where,
+      raw: true,
+      priority: [['order', 'ASC']]
```
gds-gharper, line 66: should this be `order: [['priority','ASC']]`

pages/admin/edit-category-field/EditCategoryField.html
gds-gharper: Yes, I was suggesting the absence of the parameter could be used as an indicator to create, rather than a magic string

PR #233, opened 2020-08-24
allow entities to have parents of the same entity

dtaylor7:


pages/data-entry-entity/category-template/CategoryTemplate.js
```diff
@@ -267,10 +268,16 @@ class CategoryTemplate extends Page {
...
 
+        const parentFieldNames = entity.parents.map(parent => `parent${parent.category.name}PublicId`);
+
         if(field.name === 'publicId') {
           value = entity.publicId;
-        } else if(field.name === 'parentPublicId' && entity.parents.length) {
-          value = entity.parents[0].publicId;
+        } else if(parentFieldNames.includes(field.name)) {
+          const parent = entity.parents.find(parent => {
+            return `parent${parent.category.name}PublicId` === field.name
```
gds-gharper, line 277: I see this magic string in 2 places in the code, should this be a function encapsulated within the entity model somewhere and just called there?

pages/tableau-export/TableauExport.js
```diff
@@ -28,7 +28,7 @@ class TableauExport extends Page {
     return this.req.params && this.req.params.type === 'projects';
   }
 
-  async addParents(entity, entityObject) {
+  async addParents(entity, entityObject, replaceArraysWithNumberedKeyValues = true) {
```
gds-gharper, line 31: The variable names here are pretty confusing.  If I'm reading this right entity is the actual object from the database and entityObject is a map of the fields to values.

Perhaps change entityObject to entityFieldmap (or just fieldmap)?

gds-gharper (changes requested):

gds-gharper (approved):

PR #240, opened 2020-08-28
convert user roles

AhdabNasir:
Administrators now have all roles [Admin, Uploader, Viewer, Management, Management Overview]
Uploader now has uploader and viewer role  [Uploader, Viewer]
Viewer has viewer role [Viewer]


gds-gharper:
uploader should also get viewer (as that's what they currently have)

gds-gharper (approved):
We'll probably need to clear out the user role table before this runs on staging.  Also I've just realised that we should have a compound primary key on the user_role table to prevent a user having the same role twice.

AhdabNasir:
> We'll probably need to clear out the user role table before this runs on staging. Also I've just realised that we should have a compound primary key on the user_role table to prevent a user having the same role twice.
> 
> The original design had this:
> 
> `` CREATE TABLE IF NOT EXISTS `user_role` ( `user_id` INT NOT NULL, `role_id` INT NOT NULL, PRIMARY KEY (`user_id`, `role_id`), INDEX `fk_user_role_role1_idx` (`role_id` ASC), CONSTRAINT `fk_user_role_role1` FOREIGN KEY (`role_id`) REFERENCES `role` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION, CONSTRAINT `fk_user_role_user1` FOREIGN KEY (`user_id`) REFERENCES `user` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION) ENGINE = InnoDB; ``

Ok, I'll get this merged in and will work on this on a separate branch

PR #242, opened 2020-09-01
Bpdg borders only & fix auto ID

dtaylor7:
- Error on import if the theme is borders and user permissions do not include BPDG
 - Fix auto-increment ID for entities

### DIFF
--- a/models/category.js
+++ b/models/category.js
@@ -90,7 +90,8 @@ Category.init({
   },
   currentMaxId: {
     type: INTEGER,
-    field: "current_max_id"
+    field: "current_max_id",
+    defaultValue: 0
   }
 }, { sequelize, modelName: 'category', tableName: 'category', timestamps: false });
 
--- a/models/category.js
+++ b/models/category.js
@@ -87,6 +87,10 @@ Category.init({
   publicIdFormat: {
     type: STRING(50),
     field: "public_id_format"
+  },
+  currentMaxId: {
+    type: INTEGER,
+    field: "current_max_id"
   }
 }, { sequelize, modelName: 'category', tableName: 'category', timestamps: false });
 
--- a/models/entity.js
+++ b/models/entity.js
@@ -98,9 +98,23 @@ class Entity extends Model {
       options.lock = transaction.LOCK
     }
 
-    await sequelize.query(`UPDATE category SET current_max_id = current_max_id + 1 WHERE id=${category.id}`, options);
-    const updatedCategory = await sequelize.query(`SELECT * FROM category WHERE id=${category.id} LIMIT 1 FOR UPDATE`, options);
-    return sprintf("%s%02d", category.publicIdFormat, updatedCategory[0][0].current_max_id);
+    const categories = await sequelize.query('SELECT * FROM category WHERE id=? LIMIT 1 FOR UPDATE', Object.assign({
+      mapToModel: true,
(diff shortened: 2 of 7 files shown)


models/entity.js
```diff
@@ -92,24 +92,15 @@ class Entity extends Model {
...
-    const entities = await sequelize.query(`SELECT public_id FROM entity WHERE category_id=${category.id} ORDER BY public_id DESC LIMIT 1 FOR UPDATE`, options);
-
-    if(!entities || !entities.length) {
-      return `${category.publicIdFormat}01`;
-    }
-
-    const current = parseInt(entities[0].publicId.split('-')[1]);
-    return sprintf("%s%02d", category.publicIdFormat, current+1);
+    await sequelize.query(`UPDATE category SET current_max_id = current_max_id + 1 WHERE id=${category.id}`, options);
+    const updatedCategory = await sequelize.query(`SELECT * FROM category WHERE id=${category.id} LIMIT 1 FOR UPDATE`, options);
```
gds-gharper, line 102: There's a race condition here, if the UPDATE happens from two different processes before the SELECT they'll both get the same ID and we'll skip one.  You need to SELECT for UPDATE, then manually add 1 and UPDATE to the value you set.  The FOR UPDATE will prevent another process doing the same.

gds-gharper (changes requested):

models/entity.js
dtaylor7: thanks, could you write out the query? to make sure I do it correctly?

gds-gharper:
```
options.replacements = [category.id];
options.model = Category;
options.mapToModel = true;
// Need to start a transaction here if we don't have one
const categories = await sequelize.query(`SELECT * FROM category WHERE id=? LIMIT 1 FOR UPDATE`, options);
if (categories.length > 0) {
  const newId = categories[0].current_max_id + 1;
  delete options.model;
  delete options.mapToModel;
  options.replacements = [newId,category.id];
  await sequelize.query('UPDATE category SET current_max_id=? WHERE id=?',options); // Need to check errors here and close transaction off if we created one
} else {
  // Throw error
}
```

May want to create a new options object rather than manipulating the existing one.

models/category.js
```diff
@@ -87,6 +87,10 @@ Category.init({
   publicIdFormat: {
     type: STRING(50),
     field: "public_id_format"
+  },
+  currentMaxId: {
+    type: INTEGER,
+    field: "current_max_id"
```
gds-gharper, line 93: probably want to default this to 0

models/entity.js
```diff
@@ -92,24 +92,29 @@ class Entity extends Model {
...
-    };
-
+    const options = {};
     if (transaction) {
       options.transaction = transaction;
       options.lock = transaction.LOCK
     }
 
-    const entities = await sequelize.query(`SELECT public_id FROM entity WHERE category_id=${category.id} ORDER BY public_id DESC LIMIT 1 FOR UPDATE`, options);
+    const categories = await sequelize.query('SELECT * FROM category WHERE id=? LIMIT 1 FOR UPDATE', Object.assign({
```
gds-gharper, line 101: What happens here if transaction isn't passed in?

dtaylor7: it will throw an error

gds-gharper (approved):

PR #245, opened 2020-09-01
notification role fix

AhdabNasir:
Tested and working - only those with uploader role should get notifications

### DIFF
--- a/notifications/upcomingMilestones.js
+++ b/notifications/upcomingMilestones.js
@@ -42,6 +42,7 @@ const getUsers = async (departmentName) => {
       }
     },{
       model: Role,
+      required: true,
       where: {
         name: 'uploader'
       } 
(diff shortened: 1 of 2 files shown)


notifications/upcomingMilestones.js
```diff
@@ -39,10 +40,12 @@ const getUsers = async (departmentName) => {
       where: {
         name: departmentName
       }
-    }],
-    where: {
-      role: 'uploader'
-    }
+    },{
+      model: Role,
```
dtaylor7, line 44: i think we need to add `required: true` here. it may be worth checking this emails the correct people

dtaylor7 (changes requested):

dtaylor7 (dismissed):

gds-gharper (approved):

PR #248, opened 2020-09-03
entity user permissions

dtaylor7:
- add a new table to link users to entities
 - only allow export of entities to users with permissions to those entities
 - only allow the import of entities to users with permissions to those entities
 - allow admin to export/import all entites

### DIFF
--- a/middleware/entityUserPermissions.js
+++ b/middleware/entityUserPermissions.js
@@ -17,6 +17,7 @@ const assignEntityIdsUserCanAccessToLocals = async (req, res, next) => {
   const allEntities = await Entity.findAll({
     attributes: ['publicId', 'id'],
     include: {
+      seperate: true,
       attributes: ['publicId', 'id'],
       model: Entity,
       as: 'children'
--- a/middleware/entityUserPermissions.js
+++ b/middleware/entityUserPermissions.js
@@ -1,0 +1,46 @@
+const Entity = require('models/entity');
+const User = require('models/user');
+
+const assignEntityIdsUserCanAccessToLocals = async (req, res, next) => {
+  const entitiesUserCanAccess = await Entity.findAll({
+    attributes: ['publicId', 'id'],
+    include: [{
+      model: User,
+      where: { id: req.user.id },
+    },{
+      attributes: ['publicId', 'id'],
+      model: Entity,
+      as: 'children'
+    }]
+  });
+
+  const allEntities = await Entity.findAll({
+    attributes: ['publicId', 'id'],
+    include: {
+      attributes: ['publicId', 'id'],
+      model: Entity,
+      as: 'children'
+    }
+  });
+
+  const getEntities = (entities, entity) => {
+    entities.push(entity);
+
+    entity.children.forEach(childEntity => {
+      const childEntityWithChildren = allEntities.find(entity => entity.id === childEntity.id);
+      getEntities(entities, childEntityWithChildren);
+    });
+  };
+
+  const entities = [];
+
+  for(const entity of entitiesUserCanAccess) {
+    getEntities(entities, entity);
+  }
+
+  res.locals.entitiesUserCanAccess = entities;
+
+  next();
+};
+
+module.exports = { assignEntityIdsUserCanAccessToLocals };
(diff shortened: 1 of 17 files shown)


migrations/16_entity_upload_permissions.js
```diff
@@ -0,0 +1,55 @@
+const { services } = require('config');
+const logger = require('services/logger');
+const Sequelize = require('sequelize');
+
+const up = async (query) => {
+  await query.createTable('entity_user', {
+    entity_id: {
+      type: Sequelize.DataTypes.INTEGER,
```
gds-gharper, line 8: We should use a consistent type for id's, below you're using `Sequelize.INTEGER(10).UNSIGNED` and here you're using `Sequelize.DataTypes.INTEGER`

migrations/16_entity_upload_permissions.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const down = async (query) => {
+  try {
+    await query.dropTable('entity_user');
+  } catch (error) {
+    logger.error(`Error rolling back ${error}`);
+  }
+
+  await query.bulkDelete('category', { name: 'Department' });
```
gds-gharper, line 40: Is there a unique key on the name field in category?

pages/data-entry-entity/category-template/CategoryTemplate.js
```diff
@@ -38,6 +42,12 @@ class CategoryTemplate extends Page {
   }
 
   async getRequest(req, res) {
+    const userIsAdmin = this.req.user.roles.map(role => role.name).includes('admin');
```
gds-gharper, line 45: This might be worth having as a function in the user model

isAdmin, isViewer etc.  I get the feeling it'll be used in quite a few places and it's worthwhile making it common.

gds-gharper (changes requested):

migrations/16_entity_upload_permissions.js
dtaylor7: the only reason i am using Sequelize.INTEGER(10).UNSIGNED is so that it matches user table. `Sequelize.DataTypes.INTEGER` will not work for the entity table

migrations/16_entity_upload_permissions.js
dtaylor7: yes, an id, but I cannot use this as I don't know what the id will be

gds-gharper: Ok, so long as the name field is unique then this is fine, though you may not need to use bulkDelete, it likely doesn't matter though.

migrations/16_entity_upload_permissions.js
gds-gharper: In general unless there's a reason to have negative numbers/ids you should always try to use unsigned integers for ID fields.  It does look like I missed this off the mysql workbench stuff though, my bad, we should potentially change this across the field in a future update, may be worth creating a ticket for it

middleware/entityUserPermissions.js
```diff
@@ -0,0 +1,47 @@
...
+    attributes: ['publicId', 'id'],
+    include: [{
+      model: User,
+      where: { id: req.user.id },
+    },{
+      attributes: ['publicId', 'id'],
+      model: Entity,
+      as: 'children'
+    }]
+  });
```
gds-gharper, line 15: You may want to be careful about double joins in different directions here.  worth checking the SQL and making sure it doesn't multiply.  ```DEBUG=sequelize:sql* npm run dev``` will help there (and give you query timings as well.

middleware/entityUserPermissions.js
```diff
@@ -0,0 +1,47 @@
...
+
+  const allEntities = await Entity.findAll({
+    attributes: ['publicId', 'id'],
+    include: {
+      seperate: true,
```
gds-gharper, line 20: `seperate` should be `separate`

gds-gharper (changes requested):

middleware/entityUserPermissions.js
dtaylor7: have checked the speed and its 3ms locally

middleware/entityUserPermissions.js
dtaylor7: good catch, i have removed this in this case, apparently only works on hasMany relations

middleware/entityUserPermissions.js
gds-gharper: It's mostly the number of rows I'm concerned about (speed can help diagnose this) in this case it should be fine since you're limiting it to only one user.

gds-gharper (approved):

PR #253, opened 2020-09-07
Outcomes measures column

leighjohnson-friday:
Added outcome measures column
Added new accordion as the existing component didn't work with the required structure
Still currently using the dummy data setup.  Next setup will be to remove subOutcomeData and integrate with the new subOutcomeMeasures method.


common/javascript/readiness-accordion.js
```diff
@@ -0,0 +1,133 @@
...
+
+    // Handle events
+    header.addEventListener('click', this.onSectionToggle.bind(this, $section))
+
+  }.bind(this))
+}
+
+// Set individual header attributes
+ReadinessAccordion.prototype.initHeaderAttributes = function ($headerWrapper, index) {
+  var $module = this
```
gds-gharper, line 64: Any particular reason for the $ sign here, none of the rest of the code uses it in variable names?

common/javascript/readiness-accordion.js
```diff
@@ -0,0 +1,133 @@
...
+
+  this.initSectionHeaders()
+}
+
+// Initialise section headers
+ReadinessAccordion.prototype.initSectionHeaders = function () {
+  // Loop through section headers
+  nodeListForEach(this.$sections, function ($section, i) {
+    // Set header attributes
+    var header = $section.querySelector('.' + this.sectionHeaderClass)
```
gds-gharper, line 51: Is this being passed through webpack as well?

If so we should probably be using let and const rather than declaring globals with var.

common/macros/readiness-accordion.njk
```diff
@@ -0,0 +1,72 @@
...
+        {% set heading = item.meta.name | safe %}
+        {% set isLastExpandable =  "last-level-expandable" if item.meta.isLastExpandable %}
+        {% set expanded =  "govuk-accordion__section--expanded" if item.expanded or active %}
+
+        <div class="govuk-accordion__section section-{{ color }} govuk-accordion__section-{{ id }} {{ isLastExpandable }} {{ expanded }}">
+          <div class="govuk-accordion__section-header section-header-{{ item.level }}">
+            <h2 class="govuk-accordion__section-heading govuk-heading-s">
+              <span class="govuk-accordion__section-button" id="{{ id }}-heading-{{ loop.index }}">
+                {{ heading }}
+                {% if item.level != 1 %} ({{ childCount }}){% endif %}
```
gds-gharper, line 20: Feels safer and a little more logical to do > 1 here.

common/javascript/readiness-accordion.js
```diff
@@ -0,0 +1,133 @@
...
+
+// Set individual header attributes
+ReadinessAccordion.prototype.initHeaderAttributes = function ($headerWrapper, index) {
+  var $module = this
+  var $span = $headerWrapper.querySelector('.' + this.sectionButtonClass)
+  var $heading = $headerWrapper.querySelector('.' + this.sectionHeadingClass)
+  var $summary = $headerWrapper.querySelector('.' + this.sectionSummaryClass)
+
+  // Copy existing span element to an actual button element, for improved accessibility.
+  var $button = document.createElement('button')
```
gds-gharper, line 70: Why are we doing this in dynamic javascript when it could be done in the template/macro below?

common/javascript/readiness-accordion.js
leighjohnson-friday: updated names

common/macros/readiness-accordion.njk
leighjohnson-friday: Done

common/javascript/readiness-accordion.js
leighjohnson-friday: Updated

common/javascript/readiness-accordion.js
leighjohnson-friday: Good point, removed logic and added to nunjucks template.

dtaylor7 (approved):

PR #255, opened 2020-09-08
Days countdown

AhdabNasir:


common/javascript/main.js
```diff
@@ -25,6 +25,23 @@ document.addEventListener('DOMContentLoaded', function() {
...
+  let countDown = new Date('Jan 01, 2021 00:00:00').getTime(),
+    x = setInterval(function() {    
+      let now = new Date().getTime(),
+        distance = countDown - now;
+      document.getElementById('countdown').innerHTML = `${Math.floor(distance / (day))} days`
+      if (distance < 0) {
+        clearInterval(x);
+        document.getElementById('countdown').innerHTML = '0 days'
+      }
+    }, 0)
```
leighjohnson-friday, line 44: You don't need to use setInterval here as the countdown only changes on page load.  Currently, this function will repeat on an infinite loop.
You can use a self-executing function or simply create a new function and call that on page load.

let countDown = new Date('Jan 01, 2021 00:00:00').getTime();
  
  (function(){
    let now = new Date().getTime()
    let distance = countDown - now;
    document.getElementById('countdown').innerHTML = `${Math.floor(distance / (day))} days`
    if (distance < 0) {
      document.getElementById('countdown').innerHTML = '0 days'
    }
  })();

dtaylor7: or chuck this in /core/pages/page.js:
```
get countDownTime() {
...
}
```

then in page: `{{ countDownTime }}`

or use a nunjucks globals

common/javascript/countdown.js
```diff
@@ -0,0 +1,17 @@
+export default () =>  {
+
+  const second = 1000,
+    minute = second * 60,
+    hour = minute * 60,
+    day = hour * 24;
+  
+  let countDown = new Date('Jan 01, 2021 00:00:00').getTime();
```
gds-gharper, line 8: Is it worth having this date either in the database or in config?  It's changed numerous times before.

At the very least make a function for it like getFinalTransitionDate, so that if we do want to change it in future it'll be easy to see what it was for

common/javascript/main.js
```diff
@@ -25,7 +26,9 @@ document.addEventListener('DOMContentLoaded', function() {
   if ($readinessAccordions) {
     $readinessAccordions.forEach($accordion => new ReadinessAccordion($accordion).init())
   }
-});
+
+  countdown()
```
gds-gharper, line 30: If we're not having this count down live without a page load( and I don't really see a reason to, people can reload easily , it's only once a day).  Then this should probably be done in the template, rather than in dynamic javascript

gds-gharper (changes requested):

leighjohnson-friday (approved):

gds-gharper (approved):

PR #261, opened 2020-09-09
Theme detail pages, improvements and bug fixes

dtaylor7:
- add theme details to print to the page
- fix errors with rag roll-ups
- fix errors with not being able to link to sub outcomes
- small performance tweaks

### DIFF
--- a/pages/theme/Theme.html
+++ b/pages/theme/Theme.html
@@ -6,6 +6,7 @@
 {% block pageTitle %}Theme{% endblock %}
 
 {% block content %}
+<h1>{{ data.theme.name }} {{ data.theme.color }}</h1>
 
 <div class="govuk-grid-row">
 
--- a/pages/theme/Theme.js
+++ b/pages/theme/Theme.js
@@ -16,6 +16,8 @@ const logger = require('services/logger');
 const groupBy = require('lodash/groupBy');
 const moment = require('moment');
 
+const rags = ['red', 'amber', 'green'];
+
 class Theme extends Page {
   static get isEnabled() {
     return config.features.transitionReadinessTheme;
@@ -60,8 +62,6 @@ class Theme extends Page {
   }
 
   applyRagRollups(entity) {
-    const rags = ['red', 'amber', 'green'];
-
     let color = '';
 
     if(entity.hmgConfidence) {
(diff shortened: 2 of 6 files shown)


pages/theme/Theme.js
```diff
@@ -315,112 +320,109 @@ class Theme extends Page {
...
       return entity.isLastExpandable = true;
+    } else if(entity.children) {
+      entity.children.forEach(this.applyUIFlags.bind(this));
     }
-    entity.children.map(this.applyUIFlags.bind(this));
   }
 
   groupByMetricId(entity) {
-    if(entity.children && entity.children[0].metricID) {
+    if(entity.children && entity.children.length && entity.children[0].metricID) {
```
leighjohnson-friday, line 331: Just thinking, would there ever be a scenario where children[0] doesn't contain metricID but children[1] does.  If so would that cause any issues?  It might not be possible, just want to double check.

pages/theme/Theme.js
```diff
@@ -315,112 +320,109 @@ class Theme extends Page {
...
-
-      // remove any categories without children
-      datas = datas.filter(data => data.children && data.children.length);
-
-      // remove any entities that have more than one parent
-      datas = datas.filter(data => data.parents.length <= 1);
+      // remove entites with no children and has one parent or less
+      entities = entities.filter(entity => {
+        const hasChildren = entity.children && entity.children.length;
+        const hasOnlyParentOrLess = entity.parents.length <= 1;
```
leighjohnson-friday, line 379: Should entity.children.length be entity.children.length > 0 incase the array was empty?

pages/theme/Theme.js
dtaylor7: no, metricID is set at the category level, so all children will be the same

pages/theme/Theme.js
dtaylor7: could be more readable, however when `entity.children.length` is 0 this is the same as false, so the test works the same as > 0

leighjohnson-friday (approved):

pages/theme/Theme.html
```diff
@@ -6,6 +6,7 @@
 {% block pageTitle %}Theme{% endblock %}
 
 {% block content %}
+<h1>{{ data.theme.name }} {{ data.theme.color }}</h1>
```
gds-gharper, line 9: Should color be text or a class on the tag?  The way it's done it largely merges into the actual title text

PR #264, opened 2020-09-09
user role updates

AhdabNasir:


gds-gharper (commented):
I'd be tempted to change the existing migration, since that's the one which has the dangerous code in the down.

AhdabNasir:
@gds-gharper When you say to change the existing migration, do you mean migration 15 which I wrote? Or the one which had the dangerous code which was migration 07?

gds-gharper:
> @gds-gharper When you say to change the existing migration, do you mean migration 15 which I wrote? Or the one which had the dangerous code which was migration 07?

07  was the migration with the dangerous code

gds-gharper (approved):

PR #267, opened 2020-09-10
Theme page iframe urls

dtaylor7:
- added iframe urls
 - added new threshold 
 - added cache to improve page load time

### DIFF
--- a/pages/theme/Theme.js
+++ b/pages/theme/Theme.js
@@ -109,8 +109,10 @@ class Theme extends Page {
     if (entity.hasOwnProperty('hmgConfidence')) {
       if (entity.hmgConfidence == 0) {
         color = "red";
-      } else if (entity.hmgConfidence == 1 || entity.hmgConfidence == 2) {
+      } else if (entity.hmgConfidence == 1) {
         color = "amber";
+      } else if (entity.hmgConfidence == 2) {
+        color = "yellow";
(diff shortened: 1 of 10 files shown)


pages/theme/Theme.js
```diff
@@ -229,17 +281,28 @@ class Theme extends Page {
   }
 
   async createEntityHierarchy(topLevelEntityPublicId) {
+    const cached = myCache.get(`entityHierarchy-${topLevelEntityPublicId}`);
+    if(cached) {
+      return JSON.parse(cached);
+    }
```
gds-gharper, line 287: This will very much need to be changed when we add permissions on entities.

services/tableau.js
```diff
@@ -3,6 +3,9 @@ const config = require('config');
 
 const getTableauUrl = async (user, workbook, view) => {
   if(!config.services.tableau.url) {
+    if(config.env === 'development') {
+      return `https://some-tableau-url.com/trusted/some-token/views/${workbook}/${view}`;
+    }
```
gds-gharper, line 8: Can't you just put this in the development conf, removes extra code

dtaylor7: we actually request a token from the tableau a little further down. as we don't have mock services yet we cant mock this out

gds-gharper (approved):

PR #272, opened 2020-09-11
viewer role update

AhdabNasir:


gds-gharper (approved):

gds-gharper:
You may want to add the viewer role to people as well on this.  because your previous migration didn't add those, I had to do it manually

PR #274, opened 2020-09-11
Speed up themes

gds-gharper:
This should fix the data multiplication issue on the themes page.  It's a little hacky but I'd have to change every instance of entity.children to entity.entityChildren.child (which is used in numerous ways and would require a lot more changes)

### DIFF
--- a/pages/theme/Theme.js
+++ b/pages/theme/Theme.js
@@ -6,6 +6,7 @@ const config = require('config');
 const Category = require('models/category');
 const CategoryField = require('models/categoryField');
 const Entity = require('models/entity');
+const EntityParent = require('models/entityParent');
 const EntityFieldEntry = require('models/entityFieldEntry');
 const Project = require('models/project');
 const Milestone = require('models/milestone');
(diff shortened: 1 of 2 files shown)


pages/theme/Theme.js
```diff
@@ -313,6 +323,30 @@ class Theme extends Page {
...
+      if (entity.entityParents.length) {
+        entity.entityParents.forEach( parentEntity => {
+          finalEntity.parents.push(parentEntity.parent);
+        });
+      }
+      finalEntity.dataValues.parents = finalEntity.parents;
+
+      allEntities.push(finalEntity);
+    });
+
```
leighjohnson-friday, line 349: This could be pulled out into a utility function in the event the logic is used elsewhere, however not need for now.

leighjohnson-friday (approved):

pages/theme/Theme.js
gds-gharper: TBH the whole section could be part of a model function but that would require even more refactorisation.  In the long run I'd just rather have any new code look at entity -> entityParent -> entity rather than use the belongsToMany code which doesn't support the separate option and there's an outstanding issue from 5 years ago on the sequelize list about it which is still active and getting responses up to now.

PR #276, opened 2020-09-11
fix rayg

dtaylor7:
- only use measures that have a filter=RAYG
- filter out measures that have filter=RAYG in tableau exports
- don't show outcome statements with no children

### DIFF
--- a/pages/tableau-export/TableauExport.js
+++ b/pages/tableau-export/TableauExport.js
@@ -122,7 +122,14 @@ class TableauExport extends Page {
       }
     });
 
-    const data = await this.getEntitiesFlatStructure(measuresCategory);
+    let data = await this.getEntitiesFlatStructure(measuresCategory);
+
+    data = data.filter(d => {
+      if (d.Filter && d.Filter === 'RAYG') {
+        return false;
+      }
+      return true;
+    });
 
     this.responseAsCSV(data, res);
   }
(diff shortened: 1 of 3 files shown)


pages/theme/Theme.js
```diff
@@ -439,18 +448,46 @@ class Theme extends Page {
...
+          }
+        }
+      };
+
+      entities.forEach(entity => {
+        if(entity.children) {
+          filterChildren(entity.children);
+        }
+      });
+
```
leighjohnson-friday, line 482: Just to check, am I correct in saying this is removing all the nested child data which isn't needed?

dtaylor7: yes

leighjohnson-friday (approved):

pages/tableau-export/TableauExport.js
```diff
@@ -122,7 +122,14 @@ class TableauExport extends Page {
...
 
-    const data = await this.getEntitiesFlatStructure(measuresCategory);
+    let data = await this.getEntitiesFlatStructure(measuresCategory);
+
+    data = data.filter(d => {
+      if (d.Filter && d.Filter === 'RAYG') {
+        return false;
+      }
+      return true;
+    });
```
gds-gharper, line 132: Is there any reason we're not just deleting this data from the database, or simply not uploading it?

PR #278, opened 2020-09-11
Restore scroll position after page reload

leighjohnson-friday:
Restores scroll position after page reload to provide a better user experience

### DIFF
--- a/common/javascript/readiness-accordion.js
+++ b/common/javascript/readiness-accordion.js
@@ -42,6 +42,8 @@ ReadinessAccordion.prototype.init = function () {
   }
 
   this.initSectionHeaders()
+  this.restoreScrollPosition()
+  window.addEventListener('scroll', this.setScrollPosition.bind(this))
 }
 
 // Initialise section headers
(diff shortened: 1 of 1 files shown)


dtaylor7 (approved):

common/javascript/readiness-accordion.js
```diff
@@ -144,4 +146,17 @@ ReadinessAccordion.prototype.setInitialState = function (section) {
...
+  }
+}
+
+ReadinessAccordion.prototype.restoreScrollPosition = function () {
+  if (this.browserSupportsSessionStorage) {
+    let scrollPosition = window.sessionStorage.getItem('scrollPosition') || 0
+    window.scrollTo(0, scrollPosition);
+  }
+}
+
```
gds-gharper, line 161: Won't this kick in even on different theme pages as well?

So you scroll on one page, click on another theme and it'll scroll on there?

It appears to discuss very similar issues to this on the issue here: https://github.com/alphagov/govuk-frontend/pull/958/files

gds-gharper (changes requested):

PR #290, opened 2020-09-17
show filters by default

AhdabNasir:
The filters on the All Data and Upcoming Milestones page will display by default

### DIFF
--- a/pages/all-data/AllData.html
+++ b/pages/all-data/AllData.html
@@ -40,7 +40,7 @@
   {{ title }}
 </h1>
 
-<div class="govuk-grid-row" id="show">
+<div class="govuk-grid-row" id="hide">
 
 <div class="govuk-grid-column-one-quarter" id="filterPanel">
   {% include "all-data/partials/filters/filters.njk" %}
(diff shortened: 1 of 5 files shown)


pages/all-data/AllData.html
```diff
@@ -40,7 +40,7 @@ <h1 class="govuk-heading-xl">
   {{ title }}
 </h1>
 
-<div class="govuk-grid-row" id="show">
+<div class="govuk-grid-row" id="hide">
```
gds-gharper, line 43: I know it was like this before, but should this be an id or a class?

AhdabNasir: It should be class as I'm using this to call it 
```
{{ govukButton({
  text: "Hide filter",
  href: "#hide"
}) }}
```

gds-gharper: is it worth changing it to that then, show/hide are pretty generic terms to use as an id (and could easily end up being used elsewhere as well)?

gds-gharper (approved):

PR #302, opened 2020-09-24
list users and reset password

dtaylor7:
Adds new admin pages to list users and allow admins reset user password

### DIFF
--- a/common/views/navigation.njk
+++ b/common/views/navigation.njk
@@ -31,7 +31,15 @@
     config.paths.admin.categoryFieldList,
     config.paths.admin.categoryField,
     config.paths.admin.categoryList,
-    config.paths.admin.headlineMeasures
+    config.paths.admin.headlineMeasures,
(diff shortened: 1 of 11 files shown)


AhdabNasir (approved):

gds-gharper (commented):
What about resetting a users 2fa without resetting the password, that's quite a common request, I'd make resetting the 2fa and resetting the password separate functions rather than the same

dtaylor7:
> What about resetting a users 2fa without resetting the password, that's quite a common request, I'd make resetting the 2fa and resetting the password separate functions rather than the same
> 
> Basically we should always reset login attempts if we need to, but we should give them the option of resetting the password or resetting the 2fa. We should also let the admin know on the list screen a users current login attempts (and display in red if blocked), if they have 2fa set at all, if they need to change their password at next login and which departments they have access to

yep that's a good point, I have added a task in story

PR #311, opened 2020-09-28
fix WDC

dtaylor7:
- ensure that all fields pull through
- use name if cant find category name

### DIFF
--- a/pages/tableau-export/TableauExport.js
+++ b/pages/tableau-export/TableauExport.js
@@ -73,15 +73,21 @@ class TableauExport extends Page {
 
       entityFieldMap[parentEntity.category.name] = entityFieldMap[parentEntity.category.name] || [];
 
-      let name = parentEntity.entityFieldEntries.find(entityFieldEntry => {
-        return entityFieldEntry.categoryField.name === 'groupDescription';
+      let nameEntry;
+      let groupDescriptionEntry;
+
+
+      parentEntity.entityFieldEntries.forEach(entityFieldEntry => {
+        if(entityFieldEntry.categoryField.name === 'groupDescription') {
+          groupDescriptionEntry = entityFieldEntry;
+        }
+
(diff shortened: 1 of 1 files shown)


pages/tableau-export/TableauExport.js
```diff
@@ -73,10 +73,16 @@ class TableauExport extends Page {
...
+      let name = parentEntity.entityFieldEntries.find(entityFieldEntry => {
         return entityFieldEntry.categoryField.name === 'groupDescription';
       });
 
+      if(!name) {
+        name = parentEntity.entityFieldEntries.find(entityFieldEntry => {
+          return entityFieldEntry.categoryField.name === 'name';
+        });
+      }
+
```
gds-gharper, line 85: It's relatively minor given the number of fields, but this will potentially loop through all the field entries twice if it doesn't find groupDescription.  If you have to add more it'll make it even worse.

You could loop through once and pull out both groupDescription and name (and any future fields in theory) and then just choose the most appropriate one.

Or turn the entire thing into a hash of name => value and then just pick it out based on the key.

pages/tableau-export/TableauExport.js
```diff
@@ -116,13 +122,21 @@ class TableauExport extends Page {
...
 
     for(const entity of entities) {
-      const entityFieldMap = entity.entityFieldEntries.reduce((object, entityFieldEntry) => {
-        object[entityFieldEntry.categoryField.displayName] = {
-          value: entityFieldEntry.value,
-          type: entityFieldEntry.categoryField.type
+      const entityFieldMap = categoryFields.reduce((object, categoryField) => {
+        const value = entity.entityFieldEntries.find(fieldEntry => fieldEntry.categoryFieldId == categoryField.id);
+        object[categoryField.displayName] = {
+          value: value && value.value,
```
gds-gharper, line 138: Won't this turn value into a true/false result?

gds-gharper (approved):

PR #316, opened 2020-09-29
dont append value if unit === none

dtaylor7:


AhdabNasir (dismissed):

helpers/transitionReadinessData.js
```diff
@@ -282,7 +282,10 @@ const mapEntityChildren = (allEntities, entity) => {
     entityFieldMap[entityFieldEntry.categoryField.name] = entityFieldEntry.value;
   }, {});
 
-  if(entityFieldMap.groupDescription && entityFieldMap.commentsOnly !== 'Yes') {
+  const hasGroupDescription = !!entityFieldMap.groupDescription;
+  const shouldAppendValue = entityFieldMap.unit !== 'none';
+
+  if(hasGroupDescription && shouldAppendValue) {
     if(entityFieldMap.unit === '%') {
       entityFieldMap.name = `${entityFieldMap.groupDescription}: ${entityFieldMap.value}%`;
```
gds-gharper, line 290: I think these are worth a comment, for anyone else browsing the code

dtaylor7: i thaught the names of the vars were comment enough? should I elaborate further?

gds-gharper: More why we're using magic strings like "none" and not just looking for an empty string

dtaylor7: i dont know the exact reason behind that, but I have added some comments so the developer is aware what should happen if the unit is "none"

gds-gharper (approved):

PR #319, opened 2020-09-30
Update column name to match template

leighjohnson-friday:
Update column name to match the template


dtaylor7 (approved):
may need to update tests?

PR #328, opened 2020-10-01
DDB-844 - Fix logout issue

leighjohnson-friday:
JWT cookie isn't cleared on logout.  

I believe adding the domain to the cookie options should fix the issue.   Worked locally fine without the change, so needs to be tested on staging.


dtaylor7 (commented):
nice one!

dtaylor7 (dismissed):

gds-gharper (commented):
Since the rest of the cookie code is in the jwt service, wouldn't it make sense to do this in there?

gds-gharper (approved):

PR #329, opened 2020-10-01
Dont map projects with NO milestones & hide decommissioned milestones

dtaylor7:
- Dont map projects with NO milestones
- hide decommissioned milestones

### DIFF
--- a/helpers/transitionReadinessData.js
+++ b/helpers/transitionReadinessData.js
@@ -226,22 +226,35 @@ const mapProjectsToEntities = async (entitesInHierarchy) => {
     return;
   }
 
-  const projects = await dao.getAllData(undefined, { uid: projectUids });
+  const projects = await dao.getAllData(undefined, {
+    uid: projectUids,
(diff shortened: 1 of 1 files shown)


helpers/transitionReadinessData.js
```diff
@@ -381,10 +394,10 @@ const getAllEntities = async () => {
...
 
 const createEntityHierarchy = async (category) => {
-  const cached = await redis.get(`cache-transition-overview`);
-  if(cached) {
-    return JSON.parse(cached);
-  }
+  // const cached = await redis.get(`cache-transition-overview`);
+  // if(cached) {
+  //   return JSON.parse(cached);
+  // }
```
gds-gharper, line 400: testing/debugging or actually removing it.  If so we should probably delete rather than comment out.

It also might be worth moving the cache into its own module, as such it's abstracted from redis if we want to change/remove it in future.

gds-gharper (changes requested):

helpers/transitionReadinessData.js
dtaylor7: all done

gds-gharper (approved):

PR #331, opened 2020-10-02
cron for clearing cache

dtaylor7:
- create a feature for caching so we can turn this off easily 
- reset cache at midnight

### DIFF
--- a/helpers/transitionReadinessData.js
+++ b/helpers/transitionReadinessData.js
@@ -14,6 +14,7 @@ const groupBy = require('lodash/groupBy');
 const moment = require('moment');
 const tableau = require('services/tableau');
 const redis = require('services/redis');
+const config = require('config');
 
 const rags = ['red', 'amber', 'yellow', 'green'];
 
@@ -400,7 +401,7 @@ const getAllEntities = async () => {
 
 const createEntityHierarchy = async (category) => {
   const cached = await redis.get(`cache-transition-overview`);
-  if(cached) {
+  if(cached && config.features.transitionReadinessCache) {
     return JSON.parse(cached);
   }
 
@@ -421,14 +422,14 @@ const createEntityHierarchy = async (category) => {
     entitesInHierarchy.push(entitesMapped)
   }
 
-  await redis.set(`cache-transition-overview`, JSON.stringify(entitesInHierarchy));
+  await redis.set(`cache-transition-overview`, JSON.stringify(entitesInHierarchy), moment().endOf('day').format('x'));
 
   return entitesInHierarchy;
 }
 
 const createEntityHierarchyForTheme = async (topLevelEntityPublicId) => {
   const cached = await redis.get(`cache-transition-${topLevelEntityPublicId}`);
-  if(cached) {
+  if(cached && config.features.transitionReadinessCache) {
     return JSON.parse(cached);
   }
 
@@ -445,7 +446,7 @@ const createEntityHierarchyForTheme = async (topLevelEntityPublicId) => {
   const topLevelEntityMapped = mapEntityChildren(allEntities, topLevelEntity);
   await mapProjectsToEntities(topLevelEntityMapped);
 
-  await redis.set(`cache-transition-${topLevelEntityPublicId}`, JSON.stringify(topLevelEntityMapped));
+  await redis.set(`cache-transition-${topLevelEntityPublicId}`, JSON.stringify(topLevelEntityMapped), moment().endOf('day').format('x'));
 
   return topLevelEntityMapped;
 }
--- a/helpers/transitionReadinessData.js
+++ b/helpers/transitionReadinessData.js
@@ -13,7 +13,7 @@ const logger = require('services/logger');
 const groupBy = require('lodash/groupBy');
 const moment = require('moment');
 const tableau = require('services/tableau');
-const redis = require('services/redis');
+const cache = require('services/cache');
 const config = require('config');
 
 const rags = ['red', 'amber', 'yellow', 'green'];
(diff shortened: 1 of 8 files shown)


gds-gharper (commented):
You can set an expiry on redis keys, why not just use that rather than using a cron?

dtaylor7:
> You can set an expiry on redis keys, why not just use that rather than using a cron?

is there a way to expire the cache at a specific time?

gds-gharper:
> > You can set an expiry on redis keys, why not just use that rather than using a cron?
> 
> is there a way to expire the cache at a specific time?

You can specify a number of seconds/milliseconds to expire it in.  You could always just calculate the number of seconds/milliseconds until <x> (so midnight tomorrow - now + some fudge factor) and then set it at that.  The advantage of that is it'll just happen automatically in the background with no overhead.

dtaylor7:
> > > You can set an expiry on redis keys, why not just use that rather than using a cron?
> > 
> > 
> > is there a way to expire the cache at a specific time?
> 
> You can specify a number of seconds/milliseconds to expire it in. You could always just calculate the number of seconds/milliseconds until (so midnight tomorrow - now + some fudge factor) and then set it at that. The advantage of that is it'll just happen automatically in the background with no overhead.
> 
> Edit: just looking expireat does what you want https://redis.io/commands/expireat

Yes I saw that just now too. Can't see how to do this with npm redis. is using cron really a no go?

gds-gharper:
> > > > You can set an expiry on redis keys, why not just use that rather than using a cron?
> > > 
> > > 
> > > is there a way to expire the cache at a specific time?
> > 
> > 
> > You can specify a number of seconds/milliseconds to expire it in. You could always just calculate the number of seconds/milliseconds until (so midnight tomorrow - now + some fudge factor) and then set it at that. The advantage of that is it'll just happen automatically in the background with no overhead.
> > Edit: just looking expireat does what you want https://redis.io/commands/expireat
> 
> Yes I saw that just now too. Can't see how to do this with npm redis. is using cron really a no go?

The library states it's s 1-1 mapping of commands from redis and that you should use the redis command reference for the details.  expireat has been in since 1.2 so it should support it.  have you tried just calling expireat(timestamp) ?

dtaylor7:
> > > > > You can set an expiry on redis keys, why not just use that rather than using a cron?
> > > > 
> > > > 
> > > > is there a way to expire the cache at a specific time?
> > > 
> > > 
> > > You can specify a number of seconds/milliseconds to expire it in. You could always just calculate the number of seconds/milliseconds until (so midnight tomorrow - now + some fudge factor) and then set it at that. The advantage of that is it'll just happen automatically in the background with no overhead.
> > > Edit: just looking expireat does what you want https://redis.io/commands/expireat
> > 
> > 
> > Yes I saw that just now too. Can't see how to do this with npm redis. is using cron really a no go?
> 
> The library states it's s 1-1 mapping of commands from redis and that you should use the redis command reference for the details. expireat has been in since 1.2 so it should support it. have you tried just calling expireat(timestamp) ?
> 
> Worst case scenario you could add the command yourself: https://www.npmjs.com/package/redis#clientsend_commandcommand_name-args-callback

ok, I have updated

helpers/transitionReadinessData.js
```diff
@@ -421,14 +422,14 @@ const createEntityHierarchy = async (category) => {
...
 
   return entitesInHierarchy;
 }
 
 const createEntityHierarchyForTheme = async (topLevelEntityPublicId) => {
   const cached = await redis.get(`cache-transition-${topLevelEntityPublicId}`);
-  if(cached) {
+  if(cached && config.features.transitionReadinessCache) {
     return JSON.parse(cached);
   }
```
gds-gharper, line 434: I would check that the feature is enabled before interacting with the cache/redis at all

helpers/transitionReadinessData.js
```diff
@@ -445,7 +446,7 @@ const createEntityHierarchyForTheme = async (topLevelEntityPublicId) => {
   const topLevelEntityMapped = mapEntityChildren(allEntities, topLevelEntity);
   await mapProjectsToEntities(topLevelEntityMapped);
 
-  await redis.set(`cache-transition-${topLevelEntityPublicId}`, JSON.stringify(topLevelEntityMapped));
+  await redis.set(`cache-transition-${topLevelEntityPublicId}`, JSON.stringify(topLevelEntityMapped), moment().endOf('day').format('x'));
```
gds-gharper, line 449: Check for the cache feature enabled before setting?

Also redis expects the unix epoch in seconds, nor milliseconds (X rather than x).

You can use `endOf('day').unix()` rather than dropping to format which should be faster too.

Please test this locally to make sure things are expiring, you can check the expiry of a key in the redis client by calling TTL <key>

As an aside, I'm not sure calling this function set is wise, I had assumed the object here was a standard redis object and that you were calling set on it, when you've actually defined your own set function with very different calling parameters to the standard.

I'd be tempted not to call it a redis object (which compounds the confusion) simply call it a cache object. cache.set("key","expiry") would be fine, it's the fact that it's called redis which makes this confusing.

If that's the case I'd be tempted to just pass in a moment object to the expiry and convert it inside the redis library, that way it's pretty agnostic if/when we ever change the mechanism

gds-gharper (changes requested):

helpers/transitionReadinessData.js
dtaylor7: done

helpers/transitionReadinessData.js
```diff
@@ -399,8 +400,8 @@ const getAllEntities = async () => {
 }
 
 const createEntityHierarchy = async (category) => {
-  const cached = await redis.get(`cache-transition-overview`);
-  if(cached) {
+  const cached = await cache.get(`cache-transition-overview`);
+  if(config.features.transitionReadinessCache && cached) {
```
gds-gharper, line 404: You're still fetching from the cache here, just throwing it away afterwards.  I would consider putting the config check inside the cache service and just returning null/false on set/get if it's disabled, that way you don't have to worry about it in any code

gds-gharper (changes requested):

helpers/transitionReadinessData.js
dtaylor7: ultimately this is only for development, but I will change again

gds-gharper (approved):

PR #332, opened 2020-10-05
get all measures for user to edit

dtaylor7:
@AhdabNasir This is the backend to support the frontend measures list

### DIFF
--- a/pages/data-entry-entity/measure-list/MeasureList.js
+++ b/pages/data-entry-entity/measure-list/MeasureList.js
@@ -1,0 +1,141 @@
+/* eslint-disable no-prototype-builtins */
+const Page = require('core/pages/page');
+const { paths } = require('config');
+const authentication = require('services/authentication');
+const { METHOD_NOT_ALLOWED } = require('http-status-codes');
+const Category = require('models/category');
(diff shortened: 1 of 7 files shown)


AhdabNasir (approved):

pages/data-entry-entity/measure-list/MeasureList.js
```diff
@@ -0,0 +1,161 @@
...
+    const measuresGroupedByGroupId = groupBy(measureEntities, entity => entity.groupID);
+    return Object.keys(measuresGroupedByGroupId).map(measureGroupDate => {
+      const measures = measuresGroupedByGroupId[measureGroupDate];
+      const measuresSorted = measures.sort((a, b) => moment(b.date, 'DD/MM/YYYY').valueOf() - moment(a.date, 'DD/MM/YYYY').valueOf());
+      return measuresSorted[0];
+    });
+  }
+
+  getRaygColour(entity) {
+    let colour;
```
leighjohnson-friday, line 128: This logic now exists in a helper function given it is used on the edit page as well, you should be able to remove this and use helper instead.

leighjohnson-friday (approved):

PR #333, opened 2020-10-05
Upcoming milestones update

AhdabNasir:
- Upcoming milestones page update
- Heading size changes
- Small All Data update


common/javascript/sort-table.js
```diff
@@ -0,0 +1,19 @@
+// Sorts table based on value
+export default () =>  {
+
+  const getCellValue = (tr, idx) => tr.children[idx].innerText || tr.children[idx].textContent;
```
gds-gharper, line 4: Is there not something off the shelf that does this?

https://www.npmjs.com/package/tablesort perhaps?

pages/upcoming-milestones/UpcomingMilestones.html
```diff
@@ -10,69 +8,85 @@
...
+      <th scope="col" class="govuk-table__header">Delivery Confidence</th>
+      <th scope="col" class="govuk-table__header">Project Impact</th>
+    </tr>
+  </thead>
+
+  {% for date in dates %}
+    <tbody class="govuk-table__body">
+        <tr class="govuk-table__row date-row">
+          <td rowspan="{{ date.totalMilestones * 2 }}" class="govuk-table__cell">
+            <p class="govuk-body govuk-!-font-weight-bold">{{ date.date }}</p>
```
gds-gharper, line 51: is `!` a valid css character?

common/javascript/sort-table.js
AhdabNasir: Ah did not see this, I'll have a look at this! Thanks Gareth. The table is built really weirdly as it has embedded rows etc as per the designs but hopefully it'll work with this NPM library. I'll have a play around.

pages/upcoming-milestones/UpcomingMilestones.html
AhdabNasir: Yes it's a GOVUK CSS class

gds-gharper: Not sure you should use ! in class names (from what I can read) but I see from the documentation for the design kit it is like  that so it's not a problem you can fix

common/javascript/sort-table.js
dtaylor7: i would say given the complex nature of our tables a custom approach is best

AhdabNasir: @gds-gharper As I expected it’s not working with the NPM library due to the way the table is built. I’ve had to create it differently so it’s not really a standard table hence why I wrote my own code for it. If we were to add filtering on other simple tables within the dashboard then this library would work fine. But as for this one and possibly all data, this will not work.

dtaylor7 (approved):

PR #334, opened 2020-10-06
DDB-881 - simple measure data page

leighjohnson-friday:
The initial page for editing measure data.  Pulls through measure data and displays a simple table.

### DIFF
--- a/common/views/navigation.njk
+++ b/common/views/navigation.njk
@@ -21,7 +21,8 @@
     config.paths.dataEntryEntity.bulkUploadFile,
     config.paths.dataEntryEntity.import,
     config.paths.dataEntryEntity.submissionSuccess,
-    config.paths.dataEntryEntity.output
+    config.paths.dataEntryEntity.output,
+    config.paths.dataEntryEntity.measureEdit
     ]
 %}
 
@@ -115,9 +116,14 @@
         Management information
         </a>
     </li>
-    <li class="govuk-header__navigation-item child-item {{ 'govuk-header__navigation-item--active' if url in addEntitiesUrls }}">
+    <li class="govuk-header__navigation-item child-item {{ 'govuk-header__navigation-item--active' if url in addEntitiesUrls and not config.paths.dataEntryEntity.measureEdit }}">
         <a class="govuk-header__link child-nav" href="{{ config.paths.dataEntryEntity.bulkUpload }}">
         Entity data import
+        </a>
+    </li>
+    <li class="govuk-header__navigation-item child-item {{ 'govuk-header__navigation-item--active' if url == config.paths.dataEntryEntity.measureEdit }}">
+        <a class="govuk-header__link child-nav" href="{{ config.paths.dataEntryEntity.measureEdit }}">
+        Measure edit
         </a>
     </li>
     </ul>
--- a/common/views/navigation.njk
+++ b/common/views/navigation.njk
@@ -21,7 +21,12 @@
     config.paths.dataEntryEntity.bulkUploadFile,
     config.paths.dataEntryEntity.import,
     config.paths.dataEntryEntity.submissionSuccess,
-    config.paths.dataEntryEntity.output,
+    config.paths.dataEntryEntity.output
+    ]
+%}
+
+{% set addMeasuresUrls = [
+    config.paths.dataEntryEntity.measureList,
     config.paths.dataEntryEntity.measureEdit
     ]
 %}
@@ -108,7 +113,7 @@
 </nav>
 {% endif %}
 
-{% if url in addProjectMilestonesUrls or url in addEntitiesUrls %}
+{% if url in addProjectMilestonesUrls or url in addEntitiesUrls  or url in addMeasuresUrls%}
 <nav class="child-navigation">
     <ul id="childNavigation" class="govuk-header__navigation" aria-label="Second Level Navigation">
     <li class="govuk-header__navigation-item child-item {{ 'govuk-header__navigation-item--active' if url in addProjectMilestonesUrls }}">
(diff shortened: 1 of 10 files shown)


common/views/navigation.njk
```diff
@@ -115,11 +116,16 @@
         Management information
         </a>
     </li>
-    <li class="govuk-header__navigation-item child-item {{ 'govuk-header__navigation-item--active' if url in addEntitiesUrls }}">
+    <li class="govuk-header__navigation-item child-item {{ 'govuk-header__navigation-item--active' if url in addEntitiesUrls and not config.paths.dataEntryEntity.measureEdit }}">
```
gds-gharper, line 119: Won't `url in addEntitiesUrls and not config.paths.dataEntryEntity.measureEdit` evaluate to (url in addEntitiesUrls) and (not config.paths.dataEntryEntity.measureEdit) which will always be false, since config.paths.dataEntryEntity.measureEdit will always be true (as it's set)?

leighjohnson-friday: You are correct.  Added a new array for measure URL's to fix the issue.

pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -0,0 +1,179 @@
...
+
+    if (!entityFieldEntries) {
+      logger.error(`EntityFieldEntry export, error finding entityFieldEntries`);
+      throw new Error(`EntityFieldEntry export, error finding entityFieldEntries`);
+    }
+
+    return entityFieldEntries;
+  }
+
+  async getCategory() {
```
dtaylor7, line 92: suggestion: this could be renamed to `getMeasureCategory` to better describe the function

pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -0,0 +1,179 @@
...
+      order: [['created_at', 'DESC']],
+      include: [{
+        model: EntityFieldEntry,
+        where: { value: this.req.params.metricId },
+        include: {
+          model: CategoryField,
+          where: { name: 'metricId' },
+        }
+      },{
+        model: Entity,
```
dtaylor7, line 59: perhaps a couple of comments here to explain why the deep joins i.e.:

```
        // get direct parent of measure i.e. outcome statement
        model: Entity,
        as: 'parents',
        include: [{
          model: Category
        }, {
        // get direct parents of outcome statement i.e. theme ( and possibly another outcome statement )
          model: Entity,
          as: 'parents',
          include: Category
        }]
```

pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -0,0 +1,179 @@
...
+    if(!userIsAdmin) {
+      const entityIdsUserCanAccess = this.res.locals.entitiesUserCanAccess.map(entity => entity.id);
+      where.id = { [Op.in]: entityIdsUserCanAccess };
+    }
+
+    const entities = await Entity.findAll({
+      where,
+      order: [['created_at', 'DESC']],
+      include: [{
+        model: EntityFieldEntry,
```
dtaylor7, line 52: you can add `separate: true` here to speed up the request

pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -0,0 +1,179 @@
...
+        }, {
+          model: Entity,
+          as: 'parents',
+          include: Category
+        }]
+      }]
+    });
+
+    for (const entity of entities) {
+      entity['entityFieldEntries'] = await this.getEntityFields(entity.id)
```
dtaylor7, line 72: you should already have the entity field entries for each entity from the above query under `entity. entityFieldEntries`

pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -0,0 +1,179 @@
...
+
+    if (!category) {
+      logger.error(`Category export, error finding Measure category`);
+      throw new Error(`Category export, error finding Measure category`);
+    }
+
+    return category;
+  }
+
+  async getMeasureTheme(parents) {
```
dtaylor7, line 107: with the query here: https://github.com/alphagov/co-eu-transition-dashboard/pull/334/files#diff-64b080c92e904557ee5fd4d9d592f928R48. you should be able to get theme for each entity, see my PR https://github.com/alphagov/co-eu-transition-dashboard/pull/332/files#diff-2dd26fe17f01dc3c66213f75d0e37e78R52

dtaylor7 (changes requested):
Looking really good, just a few comments that could possibly simplify the code a little

pages/data-entry-entity/measure-edit/MeasureEdit.js
leighjohnson-friday: Due to the where clause, we need to fetch these separately.  I tried a solution include 2 joins for the entityFieldEntries model but this had a similar issue and this seemed the best solution.

pages/data-entry-entity/measure-edit/MeasureEdit.js
leighjohnson-friday: Adding separate actually breaks the query and return other non-related entities so I've left this as it is. Think we need to  keep this as a join query.

dtaylor7: aha yes, it guess its because of the where clause

pages/data-entry-entity/measure-edit/MeasureEdit.js
dtaylor7: yep fair enough

dtaylor7 (approved):

PR #341, opened 2020-10-12
add page to clear cache

dtaylor7:


AhdabNasir (approved):
I assume this will only be hit through URL so no need for a navigation link?

PR #342, opened 2020-10-13
updated for entity upload

dtaylor7:
- import each entity one by one
- show an error if one occurs during import

### DIFF
--- a/pages/data-entry-entity/entity-import/EntityImport.js
+++ b/pages/data-entry-entity/entity-import/EntityImport.js
@@ -165,7 +165,6 @@ class EntityImport extends Page {
       await this.import(activeImport.category, entities);
       await this.removeTemporaryBulkImport(importId);
     } catch (error) {
-      this.clearData();
       logger.error(error);
       this.req.flash('Failed to import data');
       return this.res.redirect(this.url);
--- a/pages/data-entry-entity/entity-import/EntityImport.js
+++ b/pages/data-entry-entity/entity-import/EntityImport.js
@@ -40,15 +40,17 @@ class EntityImport extends Page {
     });
     const categoryFields = await Category.fieldDefinitions(categoryName);
 
-    const transaction = await sequelize.transaction();
-    try {
-      for(const entity of entities) {
+    for(const entity of entities) {
+      const transaction = await sequelize.transaction();
+      try {
+        throw new Error('error');
         await Entity.import(entity, category, categoryFields, { transaction });
+        await transaction.commit();
+      } catch (error) {
+        logger.error(`Error importing entity ${JSON.stringify(entity)}`);
+        await transaction.rollback();
+        throw error;
       }
-      await transaction.commit();
-    } catch (error) {
-      await transaction.rollback();
-      throw error;
     }
   }
 
@@ -164,6 +166,7 @@ class EntityImport extends Page {
       await this.import(activeImport.category, entities);
       await this.removeTemporaryBulkImport(importId);
     } catch (error) {
+      this.clearData();
       logger.error(error);
       this.req.flash('Failed to import data');
       return this.res.redirect(this.url);
--- a/pages/data-entry-entity/entity-import/EntityImport.js
+++ b/pages/data-entry-entity/entity-import/EntityImport.js
@@ -43,7 +43,6 @@ class EntityImport extends Page {
     for(const entity of entities) {
       const transaction = await sequelize.transaction();
       try {
-        throw new Error('error');
         await Entity.import(entity, category, categoryFields, { transaction });
         await transaction.commit();
       } catch (error) {
--- a/pages/data-entry-entity/entity-import/EntityImport.html
+++ b/pages/data-entry-entity/entity-import/EntityImport.html
@@ -16,6 +16,17 @@
 <div class="govuk-grid-row">
 <div class="govuk-grid-column-full">
 
+    {% if flash %}
+      {{ govukErrorSummary({
+        titleText: "There is a problem",
+        errorList: [
+          {
+            text: flash
+          }
+        ]
+      }) }}
+    {% endif %}
+
 {% if errors | length %}
   <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
     <h1 class="govuk-heading-l">Upload failed


pages/data-entry-entity/entity-import/EntityImport.js
```diff
@@ -40,15 +40,16 @@ class EntityImport extends Page {
...
     const categoryFields = await Category.fieldDefinitions(categoryName);
 
-    const transaction = await sequelize.transaction();
-    try {
-      for(const entity of entities) {
+    for(const entity of entities) {
+      const transaction = await sequelize.transaction();
+      try {
         await Entity.import(entity, category, categoryFields, { transaction });
+        await transaction.commit();
```
leighjohnson-friday, line 47: Will this change cause any issues if we encounter an issue halfway through an import, given each of the previous changes will have already been committed?

dtaylor7: yes, the import will stop running. any previous entities will remain in the database. 

unfortunately, the transaction times out if wrap the transaction around the entire import.

Another potential fix is to try and speed up with transaction.

AhdabNasir (approved):

pages/data-entry-entity/entity-import/EntityImport.js
gds-gharper: How is the transaction being around one or multiple queries affecting the speed?

I would expect the speed difference to be minimal, and normally wrapping multiple queries inside a transaction speeds things up rather than making them slower.  Have you looked into profiling what is taking all of the time?

dtaylor7: the problem is how long it takes for all of the updates to process, which with around 1000 mesures, it takes over 5 minutes. The transaction is timing out before it is committed in this scenario.

dtaylor7: i am, however, reluctant to merge this as although this fix allows the node application to continue to import the entities. The HTTP request times out, so the user is unaware that the process is continuing and will not know if the process has completed or failed.

It also means that if one of the entities fails to import the ones before will stay in the database which doesn't feel right.

gds-gharper: transactions don't tend to time out, what I suspect is happening is the html connection is timing out and that the node server is then forcibly closing the child (since the client has disconnected it sees no point in continuing to process the request).

What this change does I suspect is that it would fail at the same point and you'd just end up with the first x records processed and saved, and the rest not.  I'm guessing that it's working differently in development mode and whilst the http request is failing the backend is continuing to process.

There are a couple ways of handling this.

1. Speed up the transaction, find out what is taking so much time and optimise it (not always possible but in this case I suspect it is, it would be worthwhile seeing what the output of 0x is so we can see exactly what is taking the time).
2. Don't process the request live, instead accept the request, send it into a queue and have a worker process it in its own time (without any timeouts), the client can then be told its request has been submitted and check on the progress of the job.  This method works even if you can't optimise the actual transaction (due to the data size or the calculations required).  It does however require more engineering as you need to implement a queuing process and also handle that the actual transaction hasn't finished on the user side.  It does mean you can easily add spinners/progress meters though to indicate how far along it is.

dtaylor7: No, node continues the process even after the browser times out. And we eventually see an error from MySQL `ER_LOCK_WAIT_TIMEOUT`

dtaylor7: to your first point, yes ideally we need to optimize the way the data is imported.

gds-gharper: Aha, I suspect what's happening is that you're deadlocking within the transaction itself.  So you're trying to obtain an exclusive lock on the same rows twice which is causing the timeout.

Basically you're locking a row (or rows) not unlocking them and then trying to. get an exclusive lock on them again. 
 That's the only way that a lock wait would timeout.  The only place I can see where we're explicitly locking a row is getting the next publicId for an entity, however select for update within the same transaction doesn't block when I'm running in a mysql client so unless this is using multiple mysql connections that shouldn't happen.

I have seen on the createUser script that it seems to use multiple connections (I get the SSL error multiple times), however I'm not sure if that's happening here but that could cause the issue.

dtaylor7: i am not sure what i am looking at here, however here are a couple of outputs running this command throughout the import:
```
mysql> show full processlist;
+------+-----------------+-----------------+-----------+---------+--------+------------------------+-----------------------+
| Id   | User            | Host            | db        | Command | Time   | State                  | Info                  |
+------+-----------------+-----------------+-----------+---------+--------+------------------------+-----------------------+
|    5 | event_scheduler | localhost       | NULL      | Daemon  | 368247 | Waiting on empty queue | NULL                  |
| 4656 | root            | localhost       | dashboard | Query   |      0 | starting               | show full processlist |
| 5459 | root            | localhost:62460 | dashboard | Sleep   |      0 |                        | NULL                  |
| 5460 | root            | localhost:62461 | dashboard | Sleep   |      0 |                        | NULL                  |
+------+-----------------+-----------------+-----------+---------+--------+------------------------+-----------------------+
4 rows in set (0.00 sec)
```
```
mysql> show full processlist;
+------+-----------------+-----------------+-----------+---------+--------+------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Id   | User            | Host            | db        | Command | Time   | State                  | Info                                                                                                                                                                                                                                                                                                                                         |
+------+-----------------+-----------------+-----------+---------+--------+------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|    5 | event_scheduler | localhost       | NULL      | Daemon  | 368259 | Waiting on empty queue | NULL                                                                                                                                                                                                                                                                                                                                         |
| 4656 | root            | localhost       | dashboard | Query   |      0 | starting               | show full processlist                                                                                                                                                                                                                                                                                                                        |
| 5459 | root            | localhost:62460 | dashboard | Execute |      0 | update                 | INSERT INTO `entity_field_entry` (`entity_id`,`category_field_id`,`value`,`created_at`,`updated_at`) VALUES (8874,22,'DIT','2020-10-19 14:55:58','2020-10-19 14:55:58') ON DUPLICATE KEY UPDATE `entity_id`=VALUES(`entity_id`), `category_field_id`=VALUES(`category_field_id`), `value`=VALUES(`value`), `updated_at`=VALUES(`updated_at`) |
| 5460 | root            | localhost:62461 | dashboard | Sleep   |      0 |                        | NULL                                                                                                                                                                                                                                                                                                                                         |
+------+-----------------+-----------------+-----------+---------+--------+------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+```

dtaylor7: with no queries running it runs ```mysql> show full processlist;
+------+-----------------+-----------+-----------+---------+--------+------------------------+-----------------------+
| Id   | User            | Host      | db        | Command | Time   | State                  | Info                  |
+------+-----------------+-----------+-----------+---------+--------+------------------------+-----------------------+
|    5 | event_scheduler | localhost | NULL      | Daemon  | 368936 | Waiting on empty queue | NULL                  |
| 4656 | root            | localhost | dashboard | Query   |      0 | starting               | show full processlist |
+------+-----------------+-----------+-----------+---------+--------+------------------------+-----------------------+```

gds-gharper: Is that with a single transaction or with multiple.  I'm mostly interested in how it looks on a single transaction.  Those processlists look fine and nothing seems blocked.

You can also use `show open tables` to show tables which have locks on them.

dtaylor7: and yes that was on a single transaction

PR #344, opened 2020-10-14
Add data entries / measures frontend

AhdabNasir:


common/macros/measure-information.njk
```diff
@@ -0,0 +1,40 @@
+{% macro measureInformation(measure, url, metricId, data) %}
```
dtaylor7, line 1: is this a common measure? i think its just for editing measures so can just put in the measure `pages/data-entry-entity/measure-edit/` dir?

dtaylor7 (approved):

common/macros/measure-information.njk
AhdabNasir: @leighjohnson-friday See above

PR #348, opened 2020-10-15
DDB 893 - Update measure information

leighjohnson-friday:
Allow users to update measure information

### DIFF
--- a/pages/data-entry-entity/measure-edit/MeasureEdit.js
+++ b/pages/data-entry-entity/measure-edit/MeasureEdit.js
@@ -399,7 +399,7 @@ class MeasureEdit extends Page {
     const URLHash = '#measure-information';
 
     if (formErrors && formErrors.length) {
-      this.req.flash(formErrors.join('<br>'));
+      this.req.flash(formErrors);
       return this.res.redirect(`${this.measureUrl}/${URLHash}`);
     }
 
@@ -455,7 +455,7 @@ class MeasureEdit extends Page {
   async addMeasureEntityData (formData) {
     const formValidationErrors = await this.validateFormData(formData);
     if (formValidationErrors.length > 0) {
-      this.req.flash(formValidationErrors.join('<br>'));
+      this.req.flash(formValidationErrors);
       return this.res.redirect(this.measureUrl);
     }
 
--- a/pages/data-entry-entity/measure-edit/MeasureEdit.js
+++ b/pages/data-entry-entity/measure-edit/MeasureEdit.js
@@ -37,7 +37,7 @@ class MeasureEdit extends Page {
   } 
 
   get pathToBind() {
-    return `${this.url}/:metricId/:type?/:successful(successful)?`;
+    return `${this.url}/:metricId/:type?`;
   }
 
   get addMeasure() {
@@ -49,7 +49,7 @@ class MeasureEdit extends Page {
   }
 
   get successfulMode() {
-    return this.req.params && this.req.params.successful;
+    return this.req.params && this.req.params.type == 'successful';
   }
 
   get middleware() {
--- a/pages/data-entry-entity/measure-edit/MeasureEdit.js
+++ b/pages/data-entry-entity/measure-edit/MeasureEdit.js
@@ -477,7 +477,7 @@ class MeasureEdit extends Page {
     });
     const categoryFields = await Category.fieldDefinitions(categoryName);
     const transaction = await sequelize.transaction();
-    let redirectURL = this.measureUrl;
+    let redirectUrl = this.measureUrl;
 
     try {
       for(const entity of entities) {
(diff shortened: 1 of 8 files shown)


pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -57,6 +65,10 @@ class MeasureEdit extends Page {
       return res.status(METHOD_NOT_ALLOWED).send('You do not have permisson to access this resource.');
     }
 
+    if (this.successfulMode) {
+      this.clearData();
```
gds-gharper, line 69: clearData appears to be very misnamed, all it does is call saveData, which saves the data in the JWT.  This has caused us quite a lot of problems in the past.  Normally I'd expect this data to be serialised in the form itself if you wanted to keep it around, rather than using some kind of homegrown persistent stash.

If we really wanted to keep it client side, using localstorage rather than the cookie (which is limited in size) would probably be safer, though I'd still prefer serialising it in the form.

If we were to implement a serialise object into form method it could be used elsewhere we use this as well.

gds-gharper (changes requested):

pages/data-entry-entity/measure-edit/MeasureEdit.js
gds-gharper: Imagine the data came in via a form post rather than a CSV submission you'd have a set of form fields which contained the data for each measure.  You can do the same with hidden fields in your confirmation page.

You don't really want the submitted data in a state form as it's not state you want to save (hence the need for explicitly clearing it), it's ephemeral data you want to disappear when someone finishes the transaction (by closing the page for instance if they decide they don't want to save).

A dirtier way of doing that would be to encode the data as serialised JSON (which is what the JWT is doing anyway)  and put that value into a hidden form field (which looking at the spec has a default max length of 500k).  You will need to make sure that's submitted as a POST otherwise you'll hit url length limits.

We need to fix the way we are encoding the data anyway as we already have problems.  Doing it in this PR can fix it then it can be re-used elsewhere when we get time to convert the other areas.  Refactoring as you go is generally considered a good idea to keep a little bit on top of technical debt.  Lots of small improvements can add up a lot over time.  A good guideline to go is that as you're working on an area of code, try to clean it up and improve it with current knowledge, even if that does increase the effort a little bit.

dtaylor7: hmmm, that is an interesting approach, not one that I have heard of before, I am not sure how you would persist the data across HTTP requests. Perhaps you are suggesting returning HTML from the POST request, apart from the obvious issue that creates you wouldn't need this hidden input anyway as you would have the user input (payload) as part of the post request.

But in any case, the gold standard is to persist temporary data in the session so I am not sure why we would want to change for a bespoke solution like you suggest.

With regard to the session issue, we should move to a session stored in a key-value store like I originally suggested when writing the session management. This is proposed in this story https://technologyprogramme.atlassian.net/browse/DDB-923. 

With regards to small changes at a time, whilst for the most part, I agree, the session and data management are dealt with by the inherited class "Page". So we can make the changes there and it will affect pages, no need to incrementally change functionally per page.

Either way, this is digressing from the objective of this PR and delaying its delivery.

leighjohnson-friday: I would agree the current solution isn't ideal however I feel this particular issue is not directly related to this PR.

In the interest of delivering the features as promised to our stakeholders, I agree with Dom and this should be moved into its own ticket and shouldn't cause a delay.

It sounds like we need to spend some time deciding on the correct approach to take to solve the issue raised, rather than throwing something together that isn't fit for purpose down the line.

leighjohnson-friday: @gds-gharper, as discussed, removed the save and clear data functions.  Instead of redirecting, the page now renderer sthe post view so we can access the required data without the need for this to be saved anywhere.

pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -37,11 +37,23 @@ class MeasureEdit extends Page {
...
+    return `${this.url}/:metricId/:type?`;
+  }
+
+  get addMeasure() {
+    return this.req.params && this.req.params.type == 'add';
+  }
+
+  get editMeasure() {
+    return this.req.params && this.req.params.type == 'edit';
   }
```
gds-gharper, line 49: Do we need to have a parameter to determine whether to add or edit a measure?

If theres an ID it should be edited, otherwise add it?

leighjohnson-friday: We use the metric id for both so this determines the required action.  The other way would be to have a hidden input in the form but I think this is nicer.

gds-gharper (approved):

PR #349, opened 2020-10-15
min width viewport

AhdabNasir:
By default it will be zoomed out

### DIFF
--- a/common/views/template.njk
+++ b/common/views/template.njk
@@ -10,6 +10,8 @@
 {% endblock %}
 
 {% block head %}
+  <meta name="viewport" content="width=1024">
+
   {% if config.services.googleAnalytics.id %}
     <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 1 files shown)


common/views/template.njk
```diff
@@ -10,6 +10,8 @@
 {% endblock %}
 
 {% block head %}
+  <meta name="viewport" content="width=1024">
```
dtaylor7, line 13: may also want `initial-scale=1`. Take it tableau fits nicely within 1024?

AhdabNasir: - When I add the` initial-scale=1`, it's zoomed in by default then the user has to zoom out. Without it, it's zoomed out by default so the user can see everything on their screen. Not sure which approach is best?

- Yes Tableu fits fine 👍

dtaylor7: yea fair enough @AhdabNasir leave it without the initial-scale

dtaylor7 (approved):

PR #356, opened 2020-10-20
DA readiness overview page / theme comment

AhdabNasir:


pages/theme/Theme.html
```diff
@@ -9,8 +9,18 @@
...
-  <h1 class="govuk-heading-l theme-{{ data.theme.color }} theme-name">{{ data.theme.name }}</h1>
-  <span class="govuk-caption-l theme-description">{{ data.theme.description }}</span>
+  <h1 class="govuk-heading-l theme-{{ data.theme.color }} theme-name">{{ data.theme.name }}
+    <span class="govuk-caption-m readiness-caption">All data is UK wide unless otherwise specified</span></h1>
+
+<div class="theme-info-margin">
+  <p class="govuk-body-l theme-description">{{ data.theme.description }}</p>
+  {% if data.theme.comment %}
+    <div class="theme-comment comment-width">
+      <p class="govuk-body">Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.</p>
```
dtaylor7, line 19: should we hide this until we have real data in?

dtaylor7: or maybe wrap in a feature flag so it doesn't get pushed to prod by accident?

dtaylor7 (changes requested):

pages/theme/Theme.html
AhdabNasir: Thank you for spotting this! I forgot to replace that with `<p class="govuk-body">{{ data.theme.comment }}</p>` just did that now

AhdabNasir: Only if there's comments in the Comment field in the Theme Category then it will display. If there's no comment then nothing will display. So not sure if a feature flag is needed?

AhdabNasir:
Closing this for now until we work on this again

PR #367, opened 2020-10-26
Measure permissions

gds-gharper:
Script to create user_metrics coming

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json",
     "lines": null
   },
-  "generated_at": "2020-03-25T15:15:24Z",
+  "generated_at": "2020-10-26T03:03:30Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -53,22 +53,6 @@
         "is_verified": false,
         "line_number": 13,
         "type": "Secret Keyword"
-      }
-    ],
-    "config/default.json": [
-      {
-        "hashed_secret": "b13df69a07d0b4ae70794cc0b0ef9452455b6abe",
-        "is_secret": true,
-        "is_verified": false,
-        "line_number": 6,
-        "type": "Secret Keyword"
-      },
-      {
-        "hashed_secret": "5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8",
-        "is_secret": true,
-        "is_verified": false,
-        "line_number": 10,
-        "type": "Basic Auth Credentials"
       }
     ],
     "docker-compose.yml": [
--- a/helpers/filterMetrics.js
+++ b/helpers/filterMetrics.js
@@ -12,11 +12,7 @@ const filterMetrics = async (user,metrics) => {
         }
       });
 
-      if (fieldEntry && metricMap[fieldEntry.value]) {
-        return true;
-      }
-
-      return false;
+      return (fieldEntry && metricMap[fieldEntry.value]);
     });
   }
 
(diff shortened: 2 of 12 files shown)


.secrets.baseline
```diff
@@ -55,22 +55,6 @@
         "type": "Secret Keyword"
       }
     ],
-    "config/default.json": [
```
dtaylor7, line 58: should this be on git?

helpers/filterMetrics.js
```diff
@@ -0,0 +1,28 @@
...
+    const metricMap = await user.getPermittedMetricMap();
+
+    filteredMetrics = filteredMetrics.filter( entity => {
+      const fieldEntry = entity.entityFieldEntries.find( fieldEntry => {
+        if (fieldEntry.categoryField.name === 'metricID') {
+          return true;
+        }
+      });
+
+      if (fieldEntry && metricMap[fieldEntry.value]) {
```
dtaylor7, line 15: you could just do `return fieldEntry && metricMap[fieldEntry.value];`

models/user.js
```diff
@@ -39,6 +40,16 @@ class User extends Model {
     return projects;
   }
 
+  async getPermittedMetricMap() {
+    const userMetrics = await this.getUserMetrics();
+    let result = {};
```
dtaylor7, line 45: this could just be an array of metric ids i.e.
```
return userMetrics.map(userMetric => userMetric.metricId);
```

pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -124,6 +125,9 @@ class MeasureEdit extends Page {
       entity['entityFieldEntries'] = await this.getEntityFields(entity.id)
     }
 
+    entities = await filterMetricsHelper.filterMetrics(this.req.user,entities);
```
dtaylor7, line 128: try to steer away from mutating variables, would be more readable as `userEntities = ...`, this makes the code more readable

dtaylor7 (approved):
few minor suggestions but not necessary

.secrets.baseline
gds-gharper: Yes, it's a map of secrets the plugin we use currently thinks are in the repository, but aren't actually secrets.

models/user.js
gds-gharper: It's several orders of magnitude faster looking up a key in a map than it is searching through an array to find one, since we're always going to be looking for "do I have permission for this metric ID" this data structure makes vastly more sense.

pages/data-entry-entity/measure-edit/MeasureEdit.js
gds-gharper: I'm conscious of memory usage here, I'm always replacing the entire content of the variable and as such i'd rather not keep two copies of the structure around in memory until such time as the garbage collector gets rid of them.

I don't see any particular problem with mutating a variable which was created within this function specifically to prune values from it.

models/user.js
dtaylor7: i believe its the same with object or array as an array, you would lookup the same way as you do currently with the array

gds-gharper: The complexity of looking up an entry in an object is either O(1) or close to it due to hashing the keys.  The complexity of looking up an entry in an unsorted array is normally O(n), and if we were to sort the array we'd get to O(log n) using a binary search.

Given that we'll be doing lots of lookups it would be substantially slower using an array (since you'd need to search through the array each time in order to check if something matched).

gds-gharper: Oh, apologies, you were talking about creating the hash, I only saw the initial line, yeah that makes sense.

You can comment on multiple lines at a time by selecting multiple when making the comment.  I lost the context there

leighjohnson-friday (approved):

PR #374, opened 2020-10-27
Measure upload edit values backend

leighjohnson-friday:
Measure upload edit values backend
Includes frontend changes and backend work to populate form data
Still work in process to save data

### DIFF
--- a/pages/data-entry-entity/measure-value/MeasureValue.js
+++ b/pages/data-entry-entity/measure-value/MeasureValue.js
@@ -18,6 +18,10 @@ class MeasureValue extends Page {
     return `${this.url}/${this.req.params.measureId}`;
   }
 
+  get successfulMode() {
+    return this.req.params && this.req.params.successful;
+  }
+
   get middleware() {
     return [
       ...authentication.protect(['uploader']),
--- a/pages/data-entry-entity/measure-value/MeasureValue.js
+++ b/pages/data-entry-entity/measure-value/MeasureValue.js
@@ -3,20 +3,13 @@ const Page = require('core/pages/page');
 const { paths } = require('config');
 const authentication = require('services/authentication');
 const { METHOD_NOT_ALLOWED } = require('http-status-codes');
-const entityUserPermissions = require('middleware/entityUserPermissions');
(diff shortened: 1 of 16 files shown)


pages/data-entry-entity/measure-value/MeasureValue.js
```diff
@@ -3,35 +3,182 @@ const Page = require('core/pages/page');
...
+
+    if(!req.user.isAdmin && !canUserAccessMetric) {
       return res.status(METHOD_NOT_ALLOWED).send('You do not have permisson to access this resource.');
     }
 
     super.getRequest(req, res);
   }
+
+  async postRequest(req, res) {
+    const canUserAccessMetric = await this.canUserAccessMetric();
```
dtaylor7, line 53: coul move to handler same with getRequest line 53:57

dtaylor7 (approved):

dtaylor7:
great job with reuse of code here 👍

dtaylor7 (approved):

pages/data-entry-entity/measure-value/MeasureValue.js
```diff
@@ -15,27 +21,168 @@ class MeasureValue extends Page {
...
+  }
+
+  async getMeasureData() {
+    const { measuresEntities }  = await this.getMeasure();
+    measures.applyLabelToEntities(measuresEntities)
+    const uiInputs = measures.calculateUiInputs(measuresEntities);
+
+    const measuresForSelectedDate = measuresEntities.filter(measure => measure.date === this.req.params.date)
+
+    if (!moment(this.req.params.date, 'DD/MM/YYYY').isValid() || measuresForSelectedDate.length === 0) {
```
gds-gharper, line 173: We have the DD/MM/YYYY pattern in a lot of places in the code, might be worth trying to consolidate that in one place.

It's quite a bit more effort but I'd almost be tempted to force everything to a moment object as soon as ingress/egress (from user/database) so we're only ever dealing with a moment object, that would simplify the code everywhere quite a bit as we seem to call moment a lot in the code.

PR #376, opened 2020-10-27
Enable measure upload for production. 

leighjohnson-friday:
Removed edit links and disabled edit measure value page.

### DIFF
--- a/common/macros/data-entry-content.njk
+++ b/common/macros/data-entry-content.njk
@@ -27,10 +27,7 @@
                     <td>
                       <p class="govuk-body">{{ measure.unit }} : {{ measure.value }}</p>
                     </td>
-                    <td>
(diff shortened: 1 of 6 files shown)


common/macros/data-entry-content.njk
```diff
@@ -27,10 +27,7 @@
                     <td>
                       <p class="govuk-body">{{ measure.unit }} : {{ measure.value }}</p>
                     </td>
-                    <td>
-                    <a class="govuk-link govuk-!-font-size-19 edit-link" href="{{ editValueUrl }}/{{ measure.id }}">
-                      Edit
```
dtaylor7, line 32: do we not just want to hide this for prod?

dtaylor7 (approved):

PR #379, opened 2020-10-30
DDB-1021 - Measure upload - edit existing values

leighjohnson-friday:
Allow users to update existing measure values

### DIFF
--- a/helpers/measures.js
+++ b/helpers/measures.js
@@ -4,6 +4,10 @@ const CategoryField = require('models/categoryField');
 const EntityFieldEntry = require('models/entityFieldEntry');
 const logger = require('services/logger');
 const uniqWith = require('lodash/uniqWith');
+const { buildDateString } = require('helpers/utils');
+const moment = require('moment');
+const validation = require('helpers/validation');
+const parse = require('helpers/parse');
 
 const applyLabelToEntities = (entities) => {
   entities.forEach(entity => {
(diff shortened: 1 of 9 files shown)


helpers/measures.js
```diff
@@ -59,9 +63,82 @@ const getCategory = async (name) => {
   return category;
 }
 
+const getMeasureEntitiesFromGroup = (groupEntities, metricId) => {
+  const measureEntities = groupEntities.filter(entity => entity.metricID === metricId)
+  const sortedEntities = measureEntities.sort((a, b) => moment(a.date, 'DD/MM/YYYY').valueOf() - moment(b.date, 'DD/MM/YYYY').valueOf());
+  return sortedEntities;
+}
+
+const removeBlankEntityInputValues = (entityInputs) => {
```
dtaylor7, line 72: we have a rathe horrible function for this already if you want to use https://github.com/alphagov/co-eu-transition-dashboard/blob/bdeeea0193288fa71626a4fe33b13615794cc84d/helpers/utils.js#L1

pages/data-entry-entity/measure-value/MeasureValue.js
```diff
@@ -104,85 +239,97 @@ class MeasureValue extends Page {
...
       entities: {},
-      date: selecteDate.split('/')
-    }
+      date: selecteDate.split("/"),
+    };
 
-    measureForSelectedDate.forEach(measure => {
-      uiInputs.forEach(uiInput => {
-        // label and label2 are generated from applyLabelToEntities using filter, filet2, filterValue, filterValue2
+    measureForSelectedDate.forEach((measure) => {
```
dtaylor7, line 284: looks like the same stuff on line 80?

dtaylor7 (approved):
more great work on the reuse. few optional comments.

a lot of code has moved around, suggest hiten to fully re-test add and edit.

👍

helpers/measures.js
leighjohnson-friday: Ah nice, will take a look

PR #382, opened 2020-10-30
Measure update value - fix for when latest date is updated

leighjohnson-friday:
Updated the RAYG row when the latest entry (no filters and only item in the group) has it's date changed to a day in the past.

### DIFF
--- a/pages/data-entry-entity/measure-value/MeasureValue.js
+++ b/pages/data-entry-entity/measure-value/MeasureValue.js
@@ -13,6 +13,7 @@ const { buildDateString } = require("helpers/utils");
 const sequelize = require("services/sequelize");
 const logger = require("services/logger");
 const uniq = require('lodash/uniq');
+const cloneDeep = require('lodash/cloneDeep');
 
 class MeasureValue extends Page {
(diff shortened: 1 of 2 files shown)


pages/data-entry-entity/measure-value/MeasureValue.js
```diff
@@ -121,24 +122,49 @@ class MeasureValue extends Page {
...
-      // We need to set the parentStatementPublicId as the import will remove and recreate the entitiy in the entityparents table
-      raygEntities.forEach(raygEntity => {
-        newEntities.push({
-          publicId: raygEntity.publicId,
-          parentStatementPublicId: raygEntity.parentStatementPublicId,
-          date: newDate,
-          value
+    if (isOnlyMeasureInGroup && doesNotHaveFilter) {
+      let { value } = newEntities[0];
+      let LatestEntryUpdate = false;
```
dtaylor7, line 135: minor this is not camcase

leighjohnson-friday: Updated

dtaylor7 (approved):

PR #386, opened 2020-11-02
DDB-1022 - Measure update - delete values

leighjohnson-friday:
Allow a user to delete measure value(s)
update RAYG row when the latest measure value is deleted for a single measure in the group with no filters
Don't allow a user to delete a value when only one date for the measure

### DIFF
--- a/models/entity.js
+++ b/models/entity.js
@@ -123,6 +123,14 @@ class Entity extends Model {
 
     return sprintf("%s%02d", category.publicIdFormat, newMaxId);
   }
+
+  static async deleteEntity(entityId, options) {
+    await this.destroy({
+      where: {
+        id: entityId
+      }
+    }, options);
+  }
 }
 
 Entity.init({
--- a/models/entity.js
+++ b/models/entity.js
@@ -124,7 +124,7 @@ class Entity extends Model {
     return sprintf("%s%02d", category.publicIdFormat, newMaxId);
   }
 
-  static async deleteEntity(entityId, options) {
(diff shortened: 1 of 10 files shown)


models/entity.js
```diff
@@ -123,6 +123,14 @@ class Entity extends Model {
 
     return sprintf("%s%02d", category.publicIdFormat, newMaxId);
   }
+
+  static async delete(entityId, options) {
```
dtaylor7, line 127: optional: personally would rather write destroy where it needs to be executed. Adding static method delete is a bit confusing as destroy is the terminology used by sequelize and mysql.

throughout the code we execute methods such as findAll, findOne, create etc inline rather than creating static methods with arguments

pages/data-entry-entity/measure-value/MeasureValue.js
```diff
@@ -26,21 +28,36 @@ class MeasureValue extends Page {
   }
 
   get pathToBind() {
-    return `${this.url}/:metricId/:groupId/:date/:successful?`;
+    return `${this.url}/:type/:metricId/:groupId/:date/:successful(successful)?`;
```
dtaylor7, line 31: you can add strict matching to `:type` if you want i.e. `${this.url}/:type(edit|delete)/:metricId/:groupId/:date/:successful(successful)?`. adds a little documentation as code for easy reading

dtaylor7 (approved):
couple of optionals comments

good use of comments for complex areas 👍

PR #388, opened 2020-11-03
Static export

dtaylor7:
- update header for govuk design
 - add two new roles: static and skip_2fa
 - if user has skip_2fa role and ip is whitelisted skip 2fa auth
 - add static site export script
 - use node cache if no redis url

### DIFF
--- a/common/views/navigation.njk
+++ b/common/views/navigation.njk
@@ -61,7 +61,7 @@
 
 <nav class="navigation">
     <ul id="navigation" class="govuk-header__navigation" aria-label="Top Level Navigation">
-    {% if req.user.roles | includes(['viewer']) %}
+    {% if req.user.roles | includes(['viewer', 'static']) %}
     <li class="govuk-header__navigation-item {{ 'govuk-header__navigation-item--active' if url in transitionReadinessUrls }}">
         <a class="govuk-header__link nav" href="{{ config.paths.readinessOverview }}">Transition readiness</a>
     </li>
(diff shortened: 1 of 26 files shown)


scripts/staticSiteGenerator.js
```diff
@@ -0,0 +1,124 @@
...
+    },
+    simple: false,
+    resolveWithFullResponse: true
+  });
+
+  return get(response, 'headers["set-cookie"]');
+};
+
+const createTemporaryDirectory = dir => {
+  logger.info(`Creating temprary directory ${options.dir}`);
```
leighjohnson-friday, line 87: Should this function be called removeTemporaryDirector instead of createTemporaryDirectory, given it's removing the folder?

dtaylor7: good spot

leighjohnson-friday (dismissed):

common/views/navigation.njk
```diff
@@ -61,7 +61,7 @@
 
 <nav class="navigation">
     <ul id="navigation" class="govuk-header__navigation" aria-label="Top Level Navigation">
-    {% if req.user.roles | includes(['viewer']) %}
+    {% if req.user.roles | includes(['viewer', 'static']) %}
```
gds-gharper, line 64: How are we separating public static and DA static, we seem to be using the static role for both limiting the page access and removing the login/logout here.

scripts/staticSiteGenerator.js
```diff
@@ -0,0 +1,124 @@
...
+    },
+    simple: false,
+    resolveWithFullResponse: true
+  });
+
+  return get(response, 'headers["set-cookie"]');
+};
+
+const removeTemporaryDirectory = dir => {
+  logger.info(`Creating temprary directory ${options.dir}`);
```
gds-gharper, line 87: temprary/temporary

common/views/navigation.njk
dtaylor7: da isnt static is it? i thaught da members would login. also this PR doesnt do anything for da roles

gds-gharper (approved):

PR #393, opened 2020-11-05
Ddb-1096 notification of updates to measures

mateen-hussain:
* Query to get daily updated measures
** Did some refactoring to eliminate duplicate code
* Cron job to run at 3:15 PM every day
** cron job gets daily updated measures
** sends measures in email template
* Clear cache before running the cron job
* Change set cache method to set expiry to the next 5PM
* Unit tests


pages/admin/headline-measures/HeadlineMeasures.js
```diff
@@ -123,7 +123,7 @@ class HeadlineMeasures extends Page {
     });
 
     const measuresAsItems = measures.map(measure => {
-      const measureName = measure.entityFieldEntries.find(entry => {
+      let measureName = measure.entityFieldEntries.find(entry => {
```
gds-gharper, line 126: Any reason to change this to a let rather than a const, it doesn't appear to be modified.

notifications/updatedMeasures.js
```diff
@@ -0,0 +1,46 @@
...
+  ), '');
+  return formattedMeasures.toString();
+}
+
+
+const notifyUpdatedMeasures = async() => {
+  cache.clear();
+  const measureEntities = await getMeasuresUpdatedToday();
+  const emails = getEmails();
+  const formattedMsrs = formatMeasures(measureEntities);
```
gds-gharper, line 40: Very very minor, but I'm not a huge fan of abbreviations on variable names unless they're getting stupidly long

services/notify.js
```diff
@@ -36,6 +42,31 @@ const sendEmailWithTempPassword = async ({ email, userId, password }) => {
...
+  for(const email of emails) {
+    try {
+      await notifyClient.sendEmail(
+        notify.updatedMeasuresKey,
+        email,
+        {
+          personalisation: {
+            measures
+          },
+          reference: `${uuid.v4()}`
```
gds-gharper, line 59: What's this uuid getting us?

We're not logging it anywhere so essentially we're giving someone a unique key which if they sent back to us wouldn't help us.

mateen-hussain: Good point. I can use it for logging

mateen-hussain: fixed

gds-gharper (approved):

PR #399, opened 2020-11-06
Ddb 1096 notification of updates to measures

mateen-hussain:
remove duplicate measures

### DIFF
--- a/notifications/updatedMeasures.js
+++ b/notifications/updatedMeasures.js
@@ -34,6 +34,7 @@ const notifyUpdatedMeasures = async() => {
   cache.clear();
   const measureEntities = await getMeasuresUpdatedToday();
   const emails = getEmails();
+  
   await notifyServices.sendMeasuresUpdatedTodayEmail({ emails, measures: measureEntities });
 }
 
--- a/notifications/updatedMeasures.js
+++ b/notifications/updatedMeasures.js
@@ -20,7 +20,7 @@ const getMeasuresUpdatedToday = async() => {
   const formattedMeasures = measureEntities.map((curVal)=>(
     `${curVal.theme} ${curVal.metricID} ${curVal.name}`));
   
-    return uniq(formattedMeasures);
+  return uniq(formattedMeasures);
 }
 
 const getEmails = () => {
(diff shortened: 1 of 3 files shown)


leighjohnson-friday (dismissed):

notifications/updatedMeasures.js
```diff
@@ -18,7 +19,8 @@ const getMeasuresUpdatedToday = async() => {
   });
   const formattedMeasures = measureEntities.map((curVal)=>(
     `${curVal.theme} ${curVal.metricID} ${curVal.name}`));
-  return formattedMeasures;
+  
+  return uniq(formattedMeasures);
```
gds-gharper, line 23: How are we getting duplicated measures?

Wouldn't it be better to filter these on retrieval rather than after?

mateen-hussain: When we edit multiple values for the same measure, then the measures query is pulling all of the updated values. As we are returning only the measure name, this is creating duplicate records for measure name.

I'm not sure if we can remove them at retrieval without having to change the query, and this query is shared, we might end of doing lot of editing and I think the quick and simple solution is to remove duplicates in the end.

gds-gharper: I was able to get all metric ID's ordered by last updated at via 
```SELECT b.value AS metric,a.updated_at FROM entity_field_entry a INNER JOIN entity_field_entry b ON (a.entity_id = b.entity_id AND b.category_field_id=13) GROUP BY b.value ORDER BY a.updated_at DESC;```

The reason I had to do the inner join was that I wanted to get updated at for all metrics.  if you literally only want the highest entity for a measure updated since <x> it becomes simpler.

```SELECT entity.* FROM entity INNER JOIN entity_field_entry a ON (a.entity_id = entity.id) INNER JOIN entity_field_entry b ON (a.entity_id = b.entity_id AND b.category_field_id=13) GROUP BY b.value ORDER BY a.updated_at DESC;```

If you only want things updated since <x>

```SELECT entity.* FROM entity INNER JOIN entity_field_entry a ON (a.entity_id = entity.id) INNER JOIN entity_field_entry b ON (a.entity_id = b.entity_id AND b.category_field_id=13) WHERE a.updated_at > '2020-11-04' GROUP BY b.value ORDER BY a.updated_at DESC;```

leighjohnson-friday (approved):

PR #400, opened 2020-11-06
allow admin to produce a static export via admin panel

dtaylor7:
- add new admin page to allow admins to create, monitor and download static exports of the site
- add new table to track exports
- once static site is generated compress files and upload to s3
- hide ga tracking code if user has static role

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -4,7 +4,6 @@
   "cookie": {
     "expires": 3600000
   },
-  "temporaryDirectory": "./dist",
   "services": {
     "redis": {},
     "mysql": {},
--- a/config/default.json
+++ b/config/default.json
@@ -4,9 +4,11 @@
   "cookie": {
     "expires": 3600000
   },
+  "temporaryDirectory": "./dist",
   "services": {
     "redis": {},
     "mysql": {},
+    "s3": {},
     "tfa": {
       "name": "Transition Taskforce Dashboard"
     },
@@ -34,7 +36,8 @@
       "userManagementList": "/admin/manage-users-list",
       "userManagementAuthentication": "/admin/manage-users-authentication",
       "clearCache": "/admin/clear-cache",
-      "createUser": "/admin/create-user"
+      "createUser": "/admin/create-user",
+      "staticExports": "/admin/static-exports"
     },
     "allData": "/all-data",
     "filterExample": "/filter-example",
--- a/migrations/23_static_export.js
+++ b/migrations/23_static_export.js
@@ -1,0 +1,51 @@
+const { services } = require('config');
+const logger = require('services/logger');
+const Sequelize = require('sequelize');
+
+const up = async (query) => {
+  await query.createTable('static_export', {
+    id: {
+      type: Sequelize.DataTypes.INTEGER,
+      primaryKey: true,
+      allowNull: false,
+      autoIncrement: true
+    },
+    status: {
+      type: Sequelize.DataTypes.ENUM('in_progress', 'complete', 'error'),
+      defaultValue: 'in_progress',
+      allowNull: false
+    },
+    url: Sequelize.DataTypes.STRING(2083),
+    error: Sequelize.DataTypes.TEXT,
+    created_at: {
+      type: Sequelize.DataTypes.DATE,
+      defaultValue: Sequelize.literal('CURRENT_TIMESTAMP')
+    },
+    updated_at: {
+      type: Sequelize.DataTypes.DATE,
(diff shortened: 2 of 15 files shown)


config/default.json
```diff
@@ -4,9 +4,11 @@
   "cookie": {
     "expires": 3600000
   },
+  "temporaryDirectory": "./dist",
```
gds-gharper, line 7: Can we not use a real temporary directory for this, rather than a pseudo one in the root of the app.  Ideally it should also be cleaned up automatically after it's finished as well

https://www.gnu.org/software/autogen/mktemp.html
https://www.npmjs.com/package/mktemp

migrations/23_static_export.js
```diff
@@ -0,0 +1,51 @@
...
+      primaryKey: true,
+      allowNull: false,
+      autoIncrement: true
+    },
+    status: {
+      type: Sequelize.DataTypes.ENUM('in_progress', 'complete', 'error'),
+      defaultValue: 'in_progress',
+      allowNull: false
+    },
+    url: Sequelize.DataTypes.STRING(2083),
```
gds-gharper, line 18: Any particular reason for this being 2083, this looks very much like a magic number.

pages/admin/static-exports/StaticExports.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  async createStaticExport() {
+    return await StaticExport.create({ status: 'in_progress' });
+  }
+
+  generateStaticSiteExport(exportUid) {
+    return new Promise((resolve, reject) => {
+      const location = `${config.temporaryDirectory}/${exportUid}/export`;
+      exec(`node scripts/staticSiteGenerator.js --url ${config.serviceUrl} --dir ${location}`, async (error) => {
```
gds-gharper, line 86: I'm not confident this is particularly secure, if the URL contains any shell special variables they'll get interpreted, we're also spawning off a new shell and new interpreter.

Couldn't we just use fork and call a library function?  That way the script used for testing can essentially be a one line call to that library function.

pages/admin/static-exports/StaticExports.js
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    const tenMinutesAgo = moment().subtract(10, 'minutes');
+    const isStale = moment(currentExport.updated_at).isBefore(tenMinutesAgo);
+    if(isStale) {
+      currentExport.status = 'error';
+      currentExport.error = 'Timeout';
+      await currentExport.save();
+      return;
+    }
```
gds-gharper, line 172: If it's timed out do we know it's actually finished rather than hung?  Is it worth storing the machine and the pid so we can check it perhaps?

migrations/23_static_export.js
dtaylor7: its the maximum length of a url

pages/admin/static-exports/StaticExports.js
dtaylor7: yes we can do, although in our discussions i understood that you wanted to run this as a separate task.

if this is required change, i would like to raise a ticket and work on it retrospectively, i can do this this sprint

pages/admin/static-exports/StaticExports.js
dtaylor7: i am not sure why it would timeout but i thought i should cover this scenario. The only reason we would be here is if the exec command had not finished so i am not sure i can get that information.

config/default.json
dtaylor7: sure will look now

dtaylor7: can you give me path to temp folder you want to use?

dtaylor7: regarding this https://www.npmjs.com/package/mktemp, what is this for?

pages/admin/static-exports/StaticExports.js
gds-gharper: I'd certainly like to fix it, it's never a good idea to execute out to a shell if you can avoid it

config/default.json
gds-gharper: Well the first link is the OS level mktemp functionality (which will pick out the OS level temporary directory, normally /tmp) and guarantee a unique filename within it.  I assume the npm module wraps that to give you a nice interface to it, rather than having to shell out, I just looked briefly and the functionality to create a directory seemed to be there.

pages/admin/static-exports/StaticExports.js
dtaylor7: created new task DDB-1276 which ill start work on asap

config/default.json
dtaylor7: now using os temp dir

gds-gharper (approved):

PR #401, opened 2020-11-09
DDB-1262 - Edit measure link update

leighjohnson-friday:
Update to back link so it always points to measure list page to prevent redirection confusion if a user has come back from delete measure screen.


mateen-hussain (approved):

mateen-hussain:
Love short PRs :)

PR #411, opened 2020-11-11
DDB-1312 - Bug fix for updating group description 

leighjohnson-friday:
Bug fix - when updating group description ensure we also update the name field


dtaylor7 (approved):
looks good, we need to make sure its fully tested before deploy to prod tomo?

mateen-hussain (approved):

PR #412, opened 2020-11-11
Base search measures page

mateen-hussain:


pages/search-transition-readiness/SearchTransitionReadiness.js
```diff
@@ -0,0 +1,29 @@
...
+
+  get middleware() {
+    return [
+      ipWhiteList,
+      ...authentication.protect(['viewer', 'static', 'devolved_administrations'])
+    ];
+  }
+
+  static get isEnabled() {
+    return true;
```
dtaylor7, line 19: change this to false

dtaylor7: means that it cannot be accessed when pushed in staging. when developing we can remove the entire function

dtaylor7 (approved):

PR #413, opened 2020-11-12
DDB-1209 Allow entities to be filtered based on role

gds-gharper:


helpers/transitionReadinessData.js
```diff
@@ -248,24 +261,41 @@ const mapProjectsToEntities = async (entitesInHierarchy) => {
       entity.children = entity.children.filter(entity => {
         const isProject = entity.categoryId === projectsCategory.id;
         if(isProject) {
-          return entity.children && entity.children.length;
+          if (entity.children && entity.children.length) {
+            // console.log(`Adding project ${entity.publicId}`);
+            return true;
```
leighjohnson-friday, line 266: There are a few commented out console logs which should really be removed

helpers/transitionReadinessData.js
```diff
@@ -248,24 +261,41 @@ const mapProjectsToEntities = async (entitesInHierarchy) => {
...
     if(entity.categoryId === projectsCategory.id) {
       const project = projects.find(project => project.uid === entity.publicId);
       if(project) {
         mapProjectToEntity(milestoneFieldDefinitions, projectFieldDefinitions, entity, project);
         entity.isLastExpandable = true;
+      /* 
+      } else {
+        console.log(`Cannot find Project ${entity.publicId} (either deleted or missing milestones)`);
+      */
+      }
```
leighjohnson-friday, line 285: Is this else statement required?

helpers/transitionReadinessData.js
```diff
@@ -431,60 +461,101 @@ const getAllEntities = async () => {
...
+      heirarchy.push(entitiesMapped)
     }
-    entitesInHierarchy.push(entitesMapped)
+
+    await cache.set(`cache-transition-overview`, JSON.stringify(heirarchy));
   }
 
-  await cache.set(`cache-transition-overview`, JSON.stringify(entitesInHierarchy));
+  heirarchy = whitelistEntityHeirarchy(entitiesUserCanAccess,heirarchy);
```
leighjohnson-friday, line 525: Does the setting of cache above need to happen after whitelistEntityHeirarchy?

gds-gharper: Explicitly after, because the page is filtered per user and different users have different filters, as such we need to cache everything

leighjohnson-friday: actually, I think this is ok

helpers/transitionReadinessData.js
gds-gharper: It's commented out, I left them in as console.logs but they could easily just be comments (the code isn't the easiest to understand and the comments make it somewhat cleaner).

Your preference, I can either change them to comments or leave them as commented out console.logs (personally, I've said in the past that commented out code isn't a great idea, though this may be an exception (commented out debugging which acts as a comment).  Your choice, can change or not

leighjohnson-friday (approved):

PR #418, opened 2020-11-13
Theme page accessible

AhdabNasir:
- Polyfil for forEach and includes

- Link should not wrap all content for items in first column, moved link onto title, added css to make block clickable

- Active/focus state on measures, projects, statements etc

- Active item now have aria-current=“page”

- Hide arrow from screen reader

- Tableau iframe - ensure title attribute is set

### DIFF
--- a/common/macros/readiness-accordion.njk
+++ b/common/macros/readiness-accordion.njk
@@ -1,3 +1,5 @@
+{% from "attribute.njk" import ariaCurrent %}
+
 {% macro accordion(params) %}
   {% set id = params.id %}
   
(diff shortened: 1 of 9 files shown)


common/macros/readiness-accordion.njk
```diff
@@ -41,17 +43,17 @@
   <ul class="govuk-body govuk-!-font-size-16 outcome-measures">
     {% for item in items %}
       {% set raygColor = "item-complete" if item.complete == "Yes" else "item-" + item.color %}
-      <li class="outcome-measure__item {% if item.activeFrom %} item-disabled {% else %} {{ raygColor }} {% endif %} {% if item.active %} outcome-measure__item--active {% endif %}">
+      <li class="outcome-measure__item {% if item.activeFrom %} item-disabled {% else %} {{ raygColor }} {% endif %} {% if item.active %} outcome-measure__item--active {% endif %}" {{ ariaCurrent(item.active) }}>
       {% if item.activeFrom %}
```
leighjohnson-friday, line 47: ariaCurrent(item.active) needs to be an attribute of the link tag

common/views/template.njk
```diff
@@ -107,5 +107,6 @@
 {% block bodyEnd %}
   {# Run JavaScript at end of the <body>, to avoid blocking the initial render. #}
   <script type="text/javascript" src="/assets/main.js"></script>
+  <script type="text/javascript" src="https://polyfill.io/v3/polyfill.min.js?features=Array.prototype.includes%2CArray.prototype.forEach"></script>
   {% block afterScript %}{% endblock %}
```
leighjohnson-friday, line 111: Ideally, these polyfills would be generated via Webpack and core-js.  As a minimum, we should include these within our code to remove a dependency on a 3rd party.

leighjohnson-friday (approved):

PR #425, opened 2020-11-17
Use a plain whitelist instead of a tree

gds-gharper:
In order to ensure that any new milestones or projects don't automatically get shown we're reverting to a pure whitelist, which lists every entity the user should have access to.


middleware/entityUserPermissions.js
```diff
@@ -81,7 +32,7 @@ async function entitiesUserCanAccess(user) {
     });
 
     for (const role of roles) {
-      whitelist = whitelist.concat(await entitiesRoleCanAccess(role));
+      whitelist = whitelist.concat(role.roleEntities);
     }
```
leighjohnson-friday, line 36: I believe the entitiesRoleCanAccess function has been used in other areas of the code which will also need updating.

middleware/entityUserPermissions.js
```diff
@@ -81,7 +32,7 @@ async function entitiesUserCanAccess(user) {
     });
```
leighjohnson-friday, line 32: Do we still need to use `{
        separate: true,
        model: RoleEntityBlacklist
      }` in the find if we are using a pure whitelist?

middleware/entityUserPermissions.js
gds-gharper: Yes, and worse, the function I created wasn't even working (wrong model being put on the whitelist), I've sorted it.

middleware/entityUserPermissions.js
gds-gharper: I wanted to keep it in case we ever need to use the blacklist again, but after fixing the other problem it's gone anyway.

leighjohnson-friday (approved):

PR #428, opened 2020-11-18
Theme accessibility fixes

AhdabNasir:
- Fixed aria broken tags
- Accessibility fixes on search
- Margin fixes
- Missing H3 content fix
- Contrast colour fix based on Cervens new designs

### DIFF
--- a/common/macros/readiness-accordion.njk
+++ b/common/macros/readiness-accordion.njk
@@ -30,7 +30,7 @@
               </div>
             {% endif %}
           </div>
-          <div id="{{ id }}-content-{{ loop.index }}" class="govuk-accordion__section-content" aria-labelledby="{{ id }}-content-{{ loop.index }}">
+          <div id="{{ id }}-content-{{ loop.index }}" class="govuk-accordion__section-content" aria-labelledby="accordion-default-{{ id }}-heading-{{ loop.index }}">
             {{ item.content | safe }}
           </div>
         </div>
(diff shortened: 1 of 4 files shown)


common/macros/readiness-accordion.njk
```diff
@@ -30,7 +30,7 @@
               </div>
             {% endif %}
           </div>
-          <div id="{{ id }}-content-{{ loop.index }}" class="govuk-accordion__section-content" aria-labelledby="{{ id }}-heading-{{ loop.index }}">
+          <div id="{{ id }}-content-{{ loop.index }}" class="govuk-accordion__section-content" aria-labelledby="{{ id }}-content-{{ loop.index }}">
```
leighjohnson-friday, line 33: I don't think is quite right, the aria-labelledby should be pointing to the element which triggers the content to show.  I think in this case it would be the button element - accordion-default-{{ id }}-heading-{{ loop.index }}

leighjohnson-friday: Actually, I think it needs to point to the button ID - accordion-default-{{ id }}-heading-{{ loop.index }}

You can check an example here: https://design-system.service.gov.uk/components/accordion/default/index.html

AhdabNasir: Thanks Leigh, you're right! I've updated it to match the selected button.

leighjohnson-friday (approved):

PR #432, opened 2020-11-19
Accessibility page

AhdabNasir:
Took relevant content from the Accessibility Coronavirus page

### DIFF
--- a/common/views/template.njk
+++ b/common/views/template.njk
@@ -100,7 +100,12 @@
     href: config.paths.privacyNotice,
     text: "Privacy Notice"
   }), footerLinks) %}
-{% endif %}
+{% else %}
+ {% set footerLinks = (footerLinks.push ({
+    href: config.paths.accessibility,
+    text: "Accessibility"
+  }), footerLinks) %}
+{% endif %}  
 
 {{ govukFooter({
   meta: {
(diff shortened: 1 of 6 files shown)


common/views/template.njk
```diff
@@ -100,7 +100,12 @@
     href: config.paths.privacyNotice,
     text: "Privacy Notice"
   }), footerLinks) %}
-{% endif %}
+{% else %}
+ {% set footerLinks = (footerLinks.push ({
+    href: config.paths.accessibility,
+    text: "Accessibility"
+  }), footerLinks) %}
+{% endif %}
```
leighjohnson-friday, line 108: I think we should probably include the accessibility page for all versions of the site, not just the static version

pages/accessibility/Accessibility.js
```diff
@@ -0,0 +1,18 @@
...
+class Accessibility extends Page {
+  get url() {
+    return paths.accessibility;
+  }
+
+  // Accessibility page is only for static users
+  get middleware() {
+    return [
+      ...authentication.protect(['static'])
+    ];
```
leighjohnson-friday, line 14: I don't think we need to restrict this

leighjohnson-friday (approved):

PR #436, opened 2020-11-20
Ddb-129: measure search filters

mateen-hussain:


common/javascript/main.js
```diff
@@ -69,9 +71,17 @@ document.addEventListener('DOMContentLoaded', function() {
 
 // The expanded state of individual instances of the accordion component persists across page loads using sessionStorage.
 // These will be removed from the session storage so the accordions will collapse rather than stay open.
-// We want to keep session storage on the readiness theme pages to prevent accordions closing on a page refresh
+// We want to keep session storage on the readiness theme and search transition pages to prevent accordions closing on a page refresh
+let clearSessionStorage = true;
 const currentPage = document.getElementsByClassName('govuk-header__navigation-item--active');
-if(!currentPage.length || currentPage[0].innerText !== 'Transition readiness'){
-  sessionStorage.clear();
+if(currentPage && currentPage.length > 0 && currentPage[0].innerText === 'Transition readiness'){
```
dtaylor7, line 77: for easier reading you could name a var here e.g.
`const isTransitionReadinessPage = currentPage && currentPage.length > 0 && currentPage[0].innerText === 'Transition readiness';`

common/javascript/search.js
```diff
@@ -20,21 +20,34 @@ Search.prototype.init = function() {
...
-    if (results && results.length) {
-      return decodeURI(results[1]);
+    const regexp = `[\?&]${name}=([^&#]*)`; // eslint-disable-line no-useless-escape
+    const params = Array.from(window.location.href.matchAll(regexp));
+    let paramValues = [];
+    if (params && params.length) {
+      paramValues = params.map(param => {
+        return decodeURI(param[1])
+      })
+      // return decodeURI(results[1]);
```
dtaylor7, line 48: is this meant to be commended out?

dtaylor7 (dismissed):
Nice work. Just one optional comment.

As as aside and to draw upon our conversations, it would be good to decouple the filters from the data so they are configured as part of the backend. This would allow extending filters trivial. Perhaps you could create a ticket jira that can be prioritised with other members of team.

common/javascript/search.js
mateen-hussain: yup will remove it

common/javascript/main.js
mateen-hussain: yes, can do that

mateen-hussain:
> Nice work. Just one optional comment.
> 
> As as aside and to draw upon our conversations, it would be good to decouple the filters from the data so they are configured as part of the backend. This would allow extending filters trivial. Perhaps you could create a ticket jira that can be prioritised with other members of team.

agreed, will do this once this PR gets merged

dtaylor7 (approved):

PR #440, opened 2020-11-23
lock daily notifications so it doesnt run twice on prod

dtaylor7:


leighjohnson-friday:
Looks good, just one thing. Do you need to create a migration script to add the initial "daily updates notifications" name to the dashboard_locks table?  I'm not sure the update will work otherwise.

notifications/upcomingMilestones.js
```diff
@@ -7,13 +7,15 @@ const NotifyClient = require('notifications-node-client').NotifyClient;
...
 const logger = require('services/logger');
-const { v4: uuidv4 } = require('uuid');
+const locks = require('helpers/locks');
 
 const dao = new DAO({
   sequelize: sequelize
 });
 
-const getProjectsDueInNextTwoDays = async () => {
+const upcomingMilestonesNotifications = {};
```
mateen-hussain, line 16: any benefit in creating an object and exporting it than exporting the functions directly ?

notifications/upcomingMilestones.js
```diff
@@ -25,15 +27,15 @@ const getProjectsDueInNextTwoDays = async () => {
...
   return projects.reduce((departments, project) => {
     departments[project.departmentName] = departments[project.departmentName] || [];
     departments[project.departmentName].push(project);
     return departments;
   }, {});
 };
 
-const getUsers = async (departmentName) => {
+upcomingMilestonesNotifications.getUsers = async (departmentName) => {
   return await User.findAll({
```
mateen-hussain, line 39: await not needed with return

notifications/upcomingMilestones.js
```diff
@@ -45,17 +47,17 @@ const getUsers = async (departmentName) => {
...
 };
 
-const sendEmails = async (projectsByDepartment) => {
+upcomingMilestonesNotifications.sendEmails = async (projectsByDepartment) => {
   const notifyClient = new NotifyClient(config.notify.apiKey);
   const summary = { infos: [], errors: [] };
 
   for(const departmentName in projectsByDepartment) {
-    const users = await getUsers(departmentName);
+    const users = await upcomingMilestonesNotifications.getUsers(departmentName);
```
mateen-hussain, line 60: `getUsers` is a function within the same file. By creating an object for `upcomingMilestonesNotifications` we are calling it like `upcomingMilestonesNotifications.getUsers(..)` when we could have simply called `getUsers(..)`.

notifications/upcomingMilestones.js
```diff
@@ -140,34 +115,28 @@ const sendSummaryNotification = async (summary = { infos: [], errors: [] }) => {
 const notifyUpcomingMilestones = async () => {
```
mateen-hussain, line 115: Suggestion for improvement. Can we write a HOC which would accept, the lockname, the function to execute when lock acquired and function to run if acquiring lock throws an error.

notifications/upcomingMilestones.js
dtaylor7: unit testing was the issue

notifications/upcomingMilestones.js
dtaylor7: yes i struggled unit testing tho

notifications/upcomingMilestones.js
dtaylor7: > Looks good, just one thing. Do you need to create a migration script to add the initial "daily updates notifications" name to the dashboard_locks table? I'm not sure the update will work otherwise.

not needed, its a unique name to reference this lock

dtaylor7:
> Looks good, just one thing. Do you need to create a migration script to add the initial "daily updates notifications" name to the dashboard_locks table? I'm not sure the update will work otherwise.

not needed, its a unique name to reference this lock

notifications/upcomingMilestones.js
mateen-hussain: Technically when we write unit tests, we do it for publicly functions. Any private function will get executed as part of the unit test (which in turn also gives coverage for those functions). 

We shouldn't be exposing private functions, just to unit test them :D

leighjohnson-friday:
> > Looks good, just one thing. Do you need to create a migration script to add the initial "daily updates notifications" name to the dashboard_locks table? I'm not sure the update will work otherwise.
> 
> not needed, its a unique name to reference this lock

I just tested the update statement `UPDATE dashboard_locks SET guid='123' WHERE name='daily updates notifications' AND guid IS NULL` but it wasn't run on any records as I don't have 'daily updates notification' in the table

dtaylor7:
> > > Looks good, just one thing. Do you need to create a migration script to add the initial "daily updates notifications" name to the dashboard_locks table? I'm not sure the update will work otherwise.
> > 
> > 
> > not needed, its a unique name to reference this lock
> 
> I just tested the update statement `UPDATE dashboard_locks SET guid='123' WHERE name='daily updates notifications' AND guid IS NULL` but it wasn't run on any records as I don't have 'daily updates notification' in the table

bah, your completely right, thanks for double checking! sorted

notifications/upcomingMilestones.js
dtaylor7: thanks, will hold for now in priority of other bugs! good idea tho!

notifications/upcomingMilestones.js
dtaylor7: interesting article, in most instances i would agree, however if a developer were to change this line https://github.com/alphagov/co-eu-transition-dashboard/pull/440/files#diff-07d547dad22cdea41f1e71871b62df6d4aaac409e244a00e48a862f4825cdcefR49 it could easily be missed by the test.  likewise here https://github.com/alphagov/co-eu-transition-dashboard/pull/440/files#diff-07d547dad22cdea41f1e71871b62df6d4aaac409e244a00e48a862f4825cdcefR32.

after reading this article in future i would probably design the helper differently. that could allow for a simpler test.

leighjohnson-friday (approved):

PR #445, opened 2020-11-24
DDB-1422 don't show hidden milestones as blank entries

gds-gharper:
Milestones which were hidden (ie Decommissioned completion state) were showing up as blank rows.  This solves that.


leighjohnson-friday:
Look fine to me however Dom is now working on a ticket for showing projects without milestones so not sure if this may not be needed.  @dtaylor7 can you confirm?

leighjohnson-friday (approved):

PR #448, opened 2020-11-25
reverting the clear filters link change

mateen-hussain:


mateen-hussain:
Issue will be fixed in next PR

PR #451, opened 2020-11-26
show projects with no milestones within transition overview

dtaylor7:
Fixes https://technologyprogramme.atlassian.net/browse/DDB-1286


gds-gharper:
It would be useful to give us the ticket number for this

helpers/transitionReadinessData.js
```diff
@@ -273,8 +280,8 @@ const mapProjectsToEntities = async (entitiesInHierarchy) => {
     }
 
     if (entity.categoryId === projectsCategory.id) {
-      if (!entity.children || entity.children.length === 0) {
-        return false;
+      if (entity.children && entity.children.length === 0) {
+        delete entity.children;
       }
     }
```
gds-gharper, line 286: I think we can lose the productsCategory check here tbh and just delete children if it's empty.

helpers/transitionReadinessData.js
```diff
@@ -262,6 +263,12 @@ const mapProjectsToEntities = async (entitiesInHierarchy) => {
         if (!mapAllEntities(entity.children[i])) {
           toDelete.push(i);
         }
+        // Only include milestones where complete is Yes or No ( dont include decommissioned milestones )
+        if(entity.categoryId === projectsCategory.id) {
+          if(!['Yes', 'No'].includes(entity.children[i].complete)) {
+            toDelete.push(i);
+          }
+        }
```
gds-gharper, line 271: Rather than do this here, I'd suggest doing it around line 302.

You can then do  ```if (!['Yes', 'No'].includes(milestone.complete)) { return false; }```

That's where the other milestone checks are, it would be good to keep them in one place

gds-gharper (changes requested):

helpers/transitionReadinessData.js
dtaylor7: i removed as we clear our children in `mapEntityChildren`

dtaylor7:
> It would be useful to give us the ticket number for this

https://technologyprogramme.atlassian.net/browse/DDB-1286

AhdabNasir (approved):

PR #461, opened 2020-11-30
Tagging table

dtaylor7:
- add new tagging tables

### DIFF
--- a/migrations/28_add_tagging_tables.js
+++ b/migrations/28_add_tagging_tables.js
@@ -2,7 +2,7 @@ const { services } = require('config');
 const logger = require('services/logger');
 const Sequelize = require('sequelize');
 
-const up = async (query) => {
+const migrate = async (query) => {
   await query.createTable('tag', {
     id: {
       type: Sequelize.DataTypes.INTEGER,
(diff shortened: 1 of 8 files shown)


models/tagEntity.js
```diff
@@ -0,0 +1,16 @@
...
+TagEntity.init({
+  entityId: {
+    type: INTEGER,
+    field: "entity_id",
+  },
+  tagId: {
+    type: INTEGER,
+    field: "tag_id",
+  }
+}, { sequelize, modelName: 'tag', tableName: 'tag', timestamps: false });
```
mateen-hussain, line 14: `modelName` and `tableName ` can't be the same as `Tag` table, can they ?

migrations/28_add_tagging_tables.js
```diff
@@ -0,0 +1,67 @@
...
+
+  try {
+    await query.dropTable('tag');
+  } catch (error) {
+    logger.error(`Error rolling back ${error}`);
+  }
+};
+
+module.exports = {
+  up: async (query) => {
```
mateen-hussain, line 56: inconsistent pattern in defining `up` and `down` functions. Keep them them the same.

mateen-hussain (changes requested):

migrations/28_add_tagging_tables.js
dtaylor7: not sure what you mean here?

dtaylor7: actually i would rather stick with existing pattern, in many occasions the up function can be very large, for this reason keeps sense to keep execution and rollback separate

mateen-hussain (approved):

PR #472, opened 2020-12-04
Ddb 1504 calculate next update due on

mateen-hussain:


helpers/measures.js
```diff
@@ -215,6 +215,7 @@ const groupMeasures = (measures) => {
       measuresSortedByDate[0].colour = rayg.getRaygColour(measuresSortedByDate[0]);
       measuresSortedByDate[0].updatedAt = maxMeasureUpdatedAt
       measuresSortedByDate[0].updatedAtDate = (maxMeasureUpdatedAt) ? maxMeasureUpdatedAt.format('DD/MM/YYYY'): null;
+      measuresSortedByDate[0].updateDueOn = (measures.length > 0 && measures[0].updateDueOn) ? moment(measures[0].updateDueOn, 'DD/MM/YYYY').format("DD/MM/YYYY") : null;
```
leighjohnson-friday, line 218: Do you need to be using moment here if the format is already coming in as 'DD/MM/YYYY'

pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -495,17 +521,33 @@ class MeasureEdit extends Page {
...
-    const { errors, parsedEntities } = await measures.validateEntities(newEntities);
+    const newEntities = await this.createEntitiesFromClonedData(clonedEntities, formData);
+    const updateDueOn = this.calculateUpdateDueOn(
+      formData, latestmeasure.date, 
+      latestmeasure.updateDueOn, latestmeasure.frequency);
+    newEntities.forEach(e=> {
+      if (e.updateDueOn) {
+        e.updateDueOn = updateDueOn
+      }
+    });
```
leighjohnson-friday, line 532: Not sure this logic is needed here as the same is happening below on line 539 and entitiesToBeSaved is initially derived from newEntities.

pages/data-entry-entity/measure-edit/MeasureEdit.js
```diff
@@ -484,9 +496,23 @@ class MeasureEdit extends Page {
...
+      const { theme, createdAt, updatedAt, colour, ...other } = m;
+      allMeasures.push(other);
+    });
+    raygEntities.forEach(m => {
+      // eslint-disable-next-line no-unused-vars
+      const { theme, createdAt, updatedAt, colour, ...other } = m;
+      allMeasures.push(other);
+    });
+    
+    const latestmeasure = maxBy(allMeasures, m => moment(m.date, 'DD/MM/YYYY'));
```
leighjohnson-friday, line 515: measuresEntities are already sorted by date, is logic needed?

helpers/measures.js
mateen-hussain: you are right. Will fix this.

pages/data-entry-entity/measure-edit/MeasureEdit.js
mateen-hussain: I wasn't aware that it was already happening, by looking at code. Hence put it just in case if it wasn't.

pages/data-entry-entity/measure-edit/MeasureEdit.js
mateen-hussain: This is needed as the `createEntitiesFromClonedData` is cloning `updateDueOn` which is of format "DD/MM/YYYY" and it would fail at `measures.validateEntities(newEntities);` as it expects any date type to be of "YYYY-MM-DD" format.

pages/data-entry-entity/measure-edit/MeasureEdit.js
mateen-hussain: I'm inclined to leave this just to ensure I'm always using the latest date, wdyt ?

leighjohnson-friday: I'd say remove it so it keeps the logic the same as other areas of the page.  If you do a find in MeasureEdit, there are a few instances where we get the latest entity

pages/data-entry-entity/measure-edit/MeasureEdit.js
leighjohnson-friday: Ok fair enough, I think we can go with it for now as it will change once the DB work is done anyway.

leighjohnson-friday (approved):

PR #494, opened 2020-12-11
hardening to WDC

dtaylor7:


common/javascript/tableau-wdc.js
```diff
@@ -21,46 +21,50 @@ const getTableauDataType = type => {
 const formatKey = key => key.replace(/[^a-zA-Z0-9_]/g, '');
 
 (function() {
+  let data = [];
```
gds-gharper, line 24: Rather than use the global we need to use tableau.connectionData which is how we should be passing data through the stages (it's specified in the tableau docs).  I'm not sure if it allows compound objects so that's why I thought we should JSON.stringify it

gds-gharper (changes requested):

gds-gharper (approved):

gds-gharper (dismissed):

gds-gharper (approved):

PR #499, opened 2020-12-16
add deactivate user page

dtaylor7:
- add new page to allow admin to deactivate user
 - dont allow user to login if they are not active

### DIFF
--- a/pages/admin/user-management/deactivate/DeactivateUser.js
+++ b/pages/admin/user-management/deactivate/DeactivateUser.js
@@ -26,7 +26,7 @@ class DeactivateUser extends Page {
   }
 
   async getUser() {
-    return await User.findOne({
+    return User.findOne({
       where: {
         id: this.req.params.userId
       }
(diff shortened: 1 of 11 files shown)


mateen-hussain:
@dtaylor7 Need few clarifications:

- Will all new users have `is_active` flag set to True ?
- Does that mean, we will have to retrospectively set this flag to true for existing users?
- I was thinking rather than `is_active` maybe it should be `is_deactivated` as it would involve minimal changes?

mateen-hussain:
Will the deactivated user be shown on user list page ? And if so will deactivate button be disabled for the user. Maybe this is our of scope for this ticket, if so then good to create a ticket in backlog.

pages/admin/user-management/deactivate/DeactivateUser.js
```diff
@@ -0,0 +1,51 @@
...
+
+  get middleware() {
+    return [
+      ...authentication.protect(['admin']),
+      flash
+    ];
+  }
+
+  async getUser() {
+    return await User.findOne({
```
mateen-hussain, line 29: await not needed :D

dtaylor7: i have thought about this and i believe its useful to keep the await ( although not strictly needed ). The reason is that it documents that the function we are calling is a promise or async await function. If we remove this a future developer would need to need lookup the user.findOne function to see if its a promise or not

mateen-hussain (approved):

PR #513, opened 2020-12-21
DDB-1642 - Permissions - create a reusable object to generate entity parent/child data

dtaylor7:
- add new entity helper
 - implement it with entity permissions

### DIFF
--- a/helpers/entity.js
+++ b/helpers/entity.js
@@ -3,63 +3,56 @@ const RoleEntity = require('models/roleEntity');
 
 class EntityHelper {
   constructor() {
-    this.entityMap = {};
+    this.entities = this.constructEntityMap();
   }
 
-  async init() {
-    return this.constructEntityMap();
-  }
+  constructEntityMap() {
+    return Entity.findAll({
+      attributes: ['publicId', 'id'],
+      include: [
+        {
+          // separate: true,
+          attributes: ['id'],
+          model: Entity,
+          as: 'children'
+        }, {
+          // separate: true,
+          attributes: ['id'],
+          model: Entity,
+          as: 'parents'
+        }, {
+          attributes: ['roleId'],
+          model: RoleEntity
+        }
+      ]
+    }).then(entities => {
+      return entities.reduce((entities, entity) => {
+        entity.roles = entity.roleEntities.reduce((roles, role) => {
+          roles[role.roleId] = role;
+          return roles;
+        }, {});
+        delete entity.roleEntities;
 
-  async constructEntityMap() {
-    const entites = await Entity.findAll({
-      include: [{
-        model: Entity,
-        as: 'children'
-      },{
-        model: Entity,
-        as: 'parents'
-      },
-      {
-        model: RoleEntity
-      }]
+        entities[entity.id] = entity;
+
+        return entities;
+      }, {});
     });
-
-    const entityMap = entites.reduce((entityMap, entity) => {
-      entityMap[entity.id] = {
-        id: entity.id,
-        publicId: entity.publicId,
-        roles: entity.roleEntities.map(role => role.roleId),
-        parents: entity.parents.map(parent => {
-          return {
-            id: parent.id,
-            publicId: parent.publicId
-          };
-        }),
-        children: entity.children.map(child => {
-          return {
-            id: child.id,
-            publicId: child.publicId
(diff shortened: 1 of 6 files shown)


helpers/entity.js
```diff
@@ -0,0 +1,66 @@
...
+const RoleEntity = require('models/roleEntity');
+
+class EntityHelper {
+  constructor() {
+    this.entityMap = {};
+  }
+
+  async init() {
+    return this.constructEntityMap();
+  }
```
gds-gharper, line 11: Why not do this in the constructor, one less call to make for the users of the module.  You don't have to await it (because you can't have async constructors) but you can declare you want it and then when it's requested you can just `(await this.entityMap)`

helpers/entity.js
```diff
@@ -0,0 +1,66 @@
...
+
+  async constructEntityMap() {
+    const entites = await Entity.findAll({
```
gds-gharper, line 14: entites => entities

helpers/entity.js
```diff
@@ -0,0 +1,66 @@
...
+        }),
+        children: entity.children.map(child => {
+          return {
+            id: child.id,
+            publicId: child.publicId
+          };
+        })
+      };
+      return entityMap;
+    }, {});
```
gds-gharper, line 46: couldn't this just be entityMap[entity.id] = entity?  No need to recreate new objects if we don't need to.  if you're wanting to reduce the object size you can use something like 

```
  const allEntities = await Entity.findAll({
    attributes: ['publicId', 'id'],
    include: [{
      separate: true
      attributes: ['publicId', 'id'],
      model: Entity,
      as: 'children'
    },{
      separate: true,
      attributes: ['publicId', 'id'],
      model: Entity,
      as: 'parents'
    },{
      separate: true,
      model: RoleEntity,
    },]
  });

File: helpers/entity.js [hunk 4]
```diff
@@ -0,0 +1,66 @@
...
+    this.entityMap = entityMap;
+  }
+
+  get allEntities() {
+    return Object.values(this.entityMap);
+  }
+
+  entitiesWithRoles(roles) {
+    return Object.values(this.entityMap).reduce((entitiesWithRoles, entity) => {
+      const entityHasRole = roles.find(role => entity.roles.includes(role.id));
```
gds-gharper, line 57: Any time you're going to be running a find in a loop around everything it's worthwhile thinking should this be an object and reduce the complexity from O(n) to O(1)

gds-gharper (changes requested):

helpers/entity.js
dtaylor7: this would then make all other function async, that ok?

gds-gharper: Hmm, good point, but I suspect a lot of them are going to have to get extra data from the database anyways and would become async.

Since this is supposed to be a layer in between the database I'd assume most things would assume they are async functions.

pages/tableau-export/TableauExport.js
```diff
@@ -73,8 +73,10 @@ class TableauExport extends Page {
       throw new Error(`Cannot find role: ${roleName}`);
     }
 
-    const entities = await entityUserPermissions.entitiesRoleCanAccess(role);
-    return entities.map(entity => entity.id);
+    const entityHelper = new EntityHelper();
+    const entities = await entityHelper.entitiesWithRoles([role]);
+
+    return Object.keys(entities);
```
gds-gharper, line 79: I don't think this should be dealing with the map, it's exposing internals from the helper.  This should simply request entities a role can access and should get the array back.  The helper should return things largely in the format they expect, rather than having to make them transform them.

gds-gharper (changes requested):

pages/tableau-export/TableauExport.js
dtaylor7: i can do, but i would need to iterate through the array to get the array of id's, getting the keys from the object if much quicker. but ill do the array and iterate if thats what you want

gds-gharper: If you look at both places you use this, nothing wants an object, they all have to reduce it to an array.  I would simply return an array of entities (which one of those places can use directly) then in the tableau export use a map.

gds-gharper (approved):

PR #520, opened 2020-12-23
DDB-1529 - Entities - List page for all entities types

dtaylor7:
- Add new page to show categories then entities within those categories
 - Extend helpers/entity to return path to each entity
 - Extend helpers/entity to also add fields for each entity

Todo:
 - cover with tests

### DIFF
--- a/helpers/entity.js
+++ b/helpers/entity.js
@@ -54,7 +54,7 @@ class EntityHelper {
     })
       .then(entities => {
         return entities.reduce((entities, entity) => {
-          if(this.options.roles) {
+          if(this.options.roles && entity.roleEntities) {
             entity.roles = entity.roleEntities.reduce((roles, role) => {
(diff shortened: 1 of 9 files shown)


helpers/entity.js
```diff
@@ -1,56 +1,142 @@
...
+      throw new Error('Must include category in constructor');
+    }
+
+    return Object.values(await this.entities)
+      .filter(entity => categoryIds.includes(entity.category.id))
+  }
+
+  async getEntityData(entityID) {
+    const entities = await this.entities;
+    if (entities[entityID]) {
```
mateen-hussain, line 104: should `else` part return something ?

leighjohnson-friday: I wouldn't say it's needed.  We only want the object if it exists.

mateen-hussain (approved):

PR #529, opened 2021-01-05
Ddb 1641 permissions populate form from data from role_entity table

mateen-hussain:


pages/admin/permissions/Permissions.js
```diff
@@ -44,7 +44,27 @@ class Permissions extends Page {
...
     if (this.req.params.categoryId) {
-      return categories.getEntitesForCategory(this.req.params.categoryId);
+      this.entityHelper = new EntityHelper({ category: true, fields: true }); 
+      const entitiesForCategory = await this.entityHelper.entitiesInCategories(
+        [parseInt(this.req.params.categoryId)]);
+      const roleEntities = await getEntitiesForRoleId(this.req.params.roleId);
+
+      entitiesForCategory.forEach(ec => {
+        const re = roleEntities.find(re => re.entityId == ec.id);
+        if (re) {
```
leighjohnson-friday, line 54: Instead of doing a find in every loop, it may be more beneficial to convert roleEntities into an ID keyed object.  This can then be looked up with the ID in the loop.

mateen-hussain: agreed. It could be quicker.

leighjohnson-friday (approved):

PR #531, opened 2021-01-08
Ddb 1643 lookup parent has should cascade

mateen-hussain:


helpers/roleEntity.js
```diff
@@ -15,6 +15,23 @@ const getEntitiesForRoleId = async(roleId) => {
   },{})
 }
 
+const doesEntityHasParentsPermission = (roleEntities, entities) => {
+  let i;
+  for(i in entities) {
+    let j;
+    const entity = entities[i];
```
leighjohnson-friday, line 22: Very very minor but you could use `for on` instead of `for in` and make it slight cleaner,
for(const entity of entities)

pages/admin/permissions/Permissions.js
```diff
@@ -48,7 +48,9 @@ class Permissions extends Page {
       const entitiesForCategory = await this.entityHelper.entitiesInCategories(
         [parseInt(this.req.params.categoryId)]);
       const roleEntities = await getEntitiesForRoleId(this.req.params.roleId);
-      entitiesForCategory.forEach(ec => {
+      let i;
+      for (i in entitiesForCategory) {
+        const ec = entitiesForCategory[i];
```
leighjohnson-friday, line 53: Same as above, could use 'for on' instead

pages/admin/permissions/Permissions.js
```diff
@@ -60,8 +62,9 @@ class Permissions extends Page {
           ec.view = false;
           ec.shouldCascade = false;
         }
-      });
-      
+        const hierarchy = await this.entityHelper.getHierarchy(ec);
+        ec.hasParentsPermission = doesEntityHasParentsPermission(roleEntities, hierarchy);
```
leighjohnson-friday, line 66: Given you now have the hierarchy for all entities it would be worth adding this to the UI to help the user understand where the entities sit within the data.  An example can be found on the entityList page, should just be a simple copy and paste.

helpers/roleEntity.js
mateen-hussain: agreed it is cleaner 👍

pages/admin/permissions/Permissions.js
mateen-hussain: I'm assuming you meant `for of` !

pages/admin/permissions/Permissions.js
mateen-hussain: yes, Hiten suggested the same. I will do this as part of the other suggestion he has raised.

leighjohnson-friday (approved):

PR #532, opened 2021-01-08
Delete entities UI

AhdabNasir:


pages/admin/entity-delete/EntityDelete.html
```diff
@@ -0,0 +1,37 @@
...
+
+{% block pageTitle %}Delete entity{% endblock %}
+
+{% block content %}
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l govuk-!-margin-bottom-2">Delete entity {{ entity.publicId }}</h1>
+    <p class="govuk-body-l">{{ entity.entityFieldEntries[0].value }}</p>
+
```
leighjohnson-friday, line 27: Theres no guarantee [0] will be the name field.  It would be better to get the CategoryField data as well and then can you specify the name field

pages/admin/entity-delete/EntityDelete.js
```diff
@@ -0,0 +1,40 @@
...
+      },
+      include: [{
+        model: Entity,
+        as: 'children',
+        include: {
+          model: EntityFieldEntry,
+        }
+      }, {
+        model: EntityFieldEntry,
+      }]
```
leighjohnson-friday, line 35: Add ```include: {
          model: CategoryField,
        }```
to both model: EntityFieldEntry and then map the data, something like:

entity.entityFieldEntries.map(entityfieldEntry => {
      entity[entityfieldEntry.categoryField.name] = entityfieldEntry.value;
    });
    
You'd need to do the same for all the children as well

leighjohnson-friday (approved):

PR #536, opened 2021-01-08
Cypress UI missed milestone list

hitengpatel:


test/cypress/integration/HMGDelivery/TS_HMG_01_VerifyAllDepartment.js
```diff
@@ -6,7 +6,7 @@ const nav = new Navigation();
 const login = new Login();
 const hmg = new HMGDelivery();
 
-const department = 'All';
+const department = Cypress.config().department;
```
mateen-hussain, line 9: In ES6 JS there is a shorthand for this `const {department} = Cypress.config()`. You dont have to change this. Just information comment.

test/cypress/support/index.js
```diff
@@ -19,9 +19,10 @@ mysql.loadDBCommands();*/
 import './db_user'
 import './db_entity'
 import './db_project'
+import './utils'
 const dev = require('../../../config/development.json')
```
mateen-hussain, line 23: optional you dont need `../` to import config.json. The project is configured to do look from root folder. Just use

`const { serviceUrl } = require('config');`

test/cypress/support/utils.js
```diff
@@ -0,0 +1,20 @@
+const groupBy = (list, callback) => {
```
mateen-hussain, line 1: see if you can use https://lodash.com/docs/4.17.15#groupBy. We already use lodash in our project and groupBy in helpers/measures.js

mateen-hussain (changes requested):
check comments for details on requested changes.

test/cypress/integration/HMGDelivery/TS_HMG_01_VerifyAllDepartment.js
hitengpatel: Ohh. That's good to know.

test/cypress/support/index.js
hitengpatel: I did tried it earlier on but it does not work for some reason on cypress so i used ../ instead

test/cypress/support/utils.js
hitengpatel: I tested with above and it works. It helped me to use other functions from same lib. I will make changes.

test/cypress/support/index.js
mateen-hussain: Check with @leighjohnson-friday It would be as simple as adding the folder to some config. Again this is optional.

test/cypress/support/index.js
```diff
@@ -19,9 +19,10 @@ mysql.loadDBCommands();*/
 import './db_user'
 import './db_entity'
 import './db_project'
+//import './utils'
```
mateen-hussain, line 22: Delete this comment

mateen-hussain (approved):

PR #537, opened 2021-01-11
Ddb 1644 - Permissions UI - Label to show permissions are set by parents

mateen-hussain:


pages/admin/permissions/Permissions.scss
```diff
@@ -34,6 +35,9 @@
 
 #permissions-table{
     margin-top: 50px;
+    .header-first {
+       width: 150rem;
```
leighjohnson-friday, line 39: Very minor, we don't really use rems' anywhere in the codebase, to be consistent use either % or px instead.

leighjohnson-friday (approved):

PR #539, opened 2021-01-12
Delete entity and parents BE

AhdabNasir:


pages/admin/entities/entity-delete/EntityDelete.js
```diff
@@ -56,6 +63,34 @@ class EntityDelete extends Page {
...
+        as: 'parents',
+        include: {
+          model: EntityFieldEntry
+        }
+      }, {
+        model: EntityFieldEntry
+      }]
+    });
+  }
+
```
leighjohnson-friday, line 83: I don't think you need the include statement on this.  The below should work

`return Entity.destroy({
      where: {
        public_id: this.req.params.publicId
      })`

AhdabNasir: Thanks Leigh! You're right, just done that and tested :D

leighjohnson-friday (approved):

PR #540, opened 2021-01-13
Ddb 1645 save permissions form data

mateen-hussain:


helpers/roleEntity.js
```diff
@@ -27,8 +29,43 @@ const doesEntityHasParentsPermission = (roleEntities, entities) => {
   return false;
 }
 
+const deleteRoleEntities = async (roleId, entityIds, transaction ) => {
+  return RoleEntity.destroy({ 
+    where: {
+      roleId,
+      entityId: entityIds 
+    } }, { transaction });
+}
```
leighjohnson-friday, line 38: I think you have your transaction in the wrong place, I believe it needs to be outside of the curly braces to work:
await RoleEntity.destroy({
        where: {
      roleId,
      entityId: entityIds 
    },
        transaction
      });

mateen-hussain: good spot.

leighjohnson-friday (approved):

PR #552, opened 2021-01-20
middleware to return all entities with view permission

mateen-hussain:


helpers/entity.js
```diff
@@ -136,6 +136,45 @@ class EntityHelper {
     await this.buildHierarchy(entity, parents)
     return parents;
   }
+  
+  async entitiesWithViewPermission(roles) {
+    if(!this.options.roles) {
```
leighjohnson-friday, line 141: It doesn't appear you use this.options.roles in this function.  Should this be checking the roles function param instead?

middleware/entityUserPermissions.js
```diff
@@ -17,10 +16,9 @@ async function entitiesUserCanAccess(user) {
         where: { userId: user.id }
       }
     });
-
-    whitelist = await entityHelper.entitiesWithRoles(roles);
```
leighjohnson-friday, line 21: I believe there is another instance in the code where entitiesWithRoles is used.  This should also be updated.

helpers/entity.js
```diff
@@ -136,6 +136,45 @@ class EntityHelper {
...
+      where: {
+        roleId: roleIds,
+        canEdit: false
+      } });
+    let entitesWithViewRoles = {};
+    for(const vr of viewRoleEntities){
+      entitesWithViewRoles[vr.entityId] = es[vr.entityId];
+      if(vr.shouldCascade && es[vr.entityId].children.length >0) {
+        const childIds = es[vr.entityId].children.map(c => c.id);
+        for(const cid of childIds) {
```
leighjohnson-friday, line 156: Do you need the map before looping through the children.  Could you just use  for(const cid of es[vr.entityId].children) { instead?

mateen-hussain: agreed.

middleware/entityUserPermissions.js
mateen-hussain: I wasnt aware of updating, all instances. WIll update it and delete that function then ?

leighjohnson-friday (approved):

PR #559, opened 2021-01-27
Ddb 1812 cache view permissions data and enable permissions in prod

mateen-hussain:


helpers/transitionReadinessData.js
```diff
@@ -510,12 +511,19 @@ const whitelistEntityHeirarchy = async (whitelist,heirarchy) => {
   return heirarchy;
 }
 
-const createEntityHierarchy = async (entitiesUserCanAccess,category) => {
+const getRoleBasedCacheName = async (prefix, roles) => {
+  const allDataRoleId = await Role.findOne({ where: { name:'all_data' } });
```
leighjohnson-friday, line 515: No need to call the database, you can check if the all_data role is in the passed roles array

const allData = roles.map(role => role.name).includes('all_data');

mateen-hussain: agreed.

leighjohnson-friday (approved):

